PR #98, opened 2014-12-19
(TK-130) Dynamic default for max-threads set

dankreek:
On machines with a large number of cores, the default thread pool size
of 100 is too small. This commit will calculate the minimum number of
threads needed to run, and use that as the max-threads value if one
isn't already set in the configuration.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
@@ -360,12 +360,12 @@
   otherwise use the default."
   [config   :- WebserverRawConfig
    num-cpus :- schema/Int]
-  (let [calc-max-threads (max (calculate-required-threads config num-cpus)
-                              default-max-threads)]
-    (or (:max-threads config)
-        (do (log/warn "Thread pool size not configured so using a size of "
-                      calc-max-threads)
-            calc-max-threads))))
+  (or (:max-threads config)
+      (let [calc-max-threads (max (calculate-required-threads config num-cpus)
+                                  default-max-threads)]
+        (log/warn "Thread pool size not configured so using a size of "
+                  calc-max-threads)
+        calc-max-threads)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Public
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
@@ -276,7 +276,9 @@
      :port         (or (:port config) default-http-port)
      :request-header-max-size (or (:request-header-max-size config) default-request-header-size)}))
 
-(schema/defn contains-https-connector? [config]
+(schema/defn ^:always-validate
+  contains-https-connector? :- schema/Bool
+  [config :- WebserverRawConfig]
   (contains-keys? config #{:ssl-port :ssl-host}))
 
 (schema/defn ^:always-validate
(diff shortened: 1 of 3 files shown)


camlow325:
:+1: with the request for an additional "success" test for max-threads when starting an actual Jetty server.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
```diff
@@ -4,11 +4,11 @@
   (:require [clojure.test :refer :all]
             [clojure.java.io :refer [resource]]
             [me.raynes.fs :as fs]
-            [schema.core :as schema]
```
nwolfe: This made me realize that we've started to turn on schema validation in our tests.  If you wanted to it'd just be:
```clj
(:require [schema.test :as schema-test])
(use-fixtures :once schema-test/validate-schema)
```

dankreek: TO remove the need for the `^:always-validate' annotation?

nwolfe: Oh yeah, looks like all the ones you added have that.  Just throwing it out there - was thinking maybe the original author had intended to do that with that schema require but maybe not.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -305,6 +317,54 @@
...
+  determine-max-threads :- schema/Int
+  "Determine the size of the Jetty thread pool. If the size is specified in the
+  config then use that value, otherwise determine the minimum number of threads
+  needed to operate. If that number is larger than the default then use it,
+  otherwise use the default."
+  [config   :- WebserverRawConfig
+   num-cpus :- schema/Int]
+  (let [calc-max-threads (max (calculate-required-threads config num-cpus)
+                              default-max-threads)]
+    (or (:max-threads config)
```
nwolfe: Don't want to be an early optimizer, but would it be worthwhile to move the calculation to the `else` block to avoid the work if it's just going to use the `(:max-threads config)` anyway?

Not a big deal, up to you

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -260,17 +260,29 @@
...
+(schema/defn ^:always-validate
   maybe-get-http-connector :- (schema/maybe WebserverConnector)
   [config :- WebserverRawConfig]
-  (if (some #(contains? config %) #{:port :host})
+  (if (contains-http-connector? config)
     {:host         (or (:host config) default-host)
      :port         (or (:port config) default-http-port)
      :request-header-max-size (or (:request-header-max-size config) default-request-header-size)}))
 
+(schema/defn contains-https-connector? [config]
```
nwolfe: Did you mean to schematize the inputs/outputs?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -243,7 +250,11 @@
   [server :- Server
    config :- config/WebserverConnector]
   (let [request-size (:request-header-max-size config)]
-    (doto (ServerConnector. server (connection-factory request-size))
+    (doto (ServerConnector. server
+                            nil nil nil
```
camlow325: I see a constructor that allows you to specify an acceptor count, selector count, and `SslContextFactory`.  I'm not seeing one that allows you to specify acceptor count, selector count, and a ConnectionFactory other than the one where you have to also pass in the `Executor`, `Scheduler`, and `ByteBufferPool`, though.

camlow325:
:+1:

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
fpringvaldsen: Ah, I just realized I was looking at the 9.2.6 docs. It was probably added sometime between the version we're using and 9.2.6.

dankreek:
Closed in favor of https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/99

PR #100, opened 2015-01-07
(TK-82) Added `queue-max-size` setting

camlow325:
This commit adds a `queue-max-size` setting for helping control the maximum number of incoming connections that can be queued by the webserver before any new connections are dropped.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
@@ -28,6 +28,7 @@
             [ring.util.codec :as codec]
             [clojure.string :as str]
             [clojure.tools.logging :as log]
+            [puppetlabs.kitchensink.core :as ks]
             [puppetlabs.trapperkeeper.services.webserver.jetty9-config :as config]
             [schema.core :as schema]))
 
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
@@ -7,15 +7,14 @@
            (org.eclipse.jetty.util.resource Resource)
            (org.eclipse.jetty.util.thread QueuedThreadPool)
            (org.eclipse.jetty.util.ssl SslContextFactory)
-           (javax.servlet.http HttpServletRequest HttpServletResponse)
-           (java.util.concurrent Executors TimeoutException)
+           (javax.servlet.http HttpServletResponse)
+           (java.util.concurrent TimeoutException)
            (org.eclipse.jetty.servlets.gzip GzipHandler)
            (org.eclipse.jetty.servlet ServletContextHandler ServletHolder DefaultServlet)
            (org.eclipse.jetty.webapp WebAppContext)
            (java.util HashSet)
            (org.eclipse.jetty.http MimeTypes)
            (javax.servlet Servlet ServletContextListener)
-           (java.io File IOException)
            (org.eclipse.jetty.proxy ProxyServlet)
            (java.net URI)
            (java.security Security)
@@ -23,14 +22,12 @@
            (clojure.lang Atom)
            (java.lang.management ManagementFactory)
            (org.eclipse.jetty.jmx MBeanContainer)
-           (org.eclipse.jetty.util URIUtil))
+           (org.eclipse.jetty.util URIUtil BlockingArrayQueue))
 
   (:require [ring.util.servlet :as servlet]
             [ring.util.codec :as codec]
             [clojure.string :as str]
-            [clojure.set :as set]
             [clojure.tools.logging :as log]
-            [puppetlabs.kitchensink.core :as ks]
             [puppetlabs.trapperkeeper.services.webserver.jetty9-config :as config]
             [schema.core :as schema]))
 
@@ -135,10 +132,17 @@
 
 (def default-graceful-stop-timeout 60000)
 
+(def default-request-idle-timeout
+  1000)
+
 (def default-proxy-idle-timeout
   "The default number of milliseconds to wait before the proxy gives up on the
   upstream server if the :idle-timeout value isn't set in the proxy config."
   60000)
+
+(def default-queue-idle-timeout 60000)
+
+(def default-queue-min-threads 8)
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Utility Functions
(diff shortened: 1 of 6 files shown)


camlow325:
Note that this should not be merged in until after the next planned tk-jetty9 release is done - 1.1.0, I believe.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
```diff
@@ -174,3 +177,126 @@
...
+  [num-requests port]
+  (doall (for [_ (range num-requests)]
+           (future
+             (http-sync/get
+               (str "http://localhost:"
+                    port
+                    "/hello")
+               {:as :text})))))
+
+(deftest queue-max-size-tests
```
camlow325: I struggled mightily with trying to create a more "end-to-end" level test that proves out this functionality.  These tests definitely have race conditions in them - e.g., with the "sleep" statements and guesses at a sufficient number of concurrent requests that would need to be made in order to tickle the upper limit on the queue.  I haven't been able to come up with a safer set of tests that exercises the functionality in a meaningful way, though.  I ultimately think it might be better to just remove these tests altogether but wanted to get thoughts from others first.

camlow325: The fragility of these tests proved out in Travis runs.

On one Travis cell, https://travis-ci.org/puppetlabs/trapperkeeper-webserver-jetty9/jobs/46255778,  the tests failed because the excessive number of http clients being created by the test caused Java OutOfMemoryErrors.  These may have been possible to overcome by bumping up the Java heap when running tests or manipulating `max-threads` on the Jetty server to be lower such that the number of concurrent requests which needed to be run were much lower.

On another Travis cell, https://travis-ci.org/puppetlabs/trapperkeeper-webserver-jetty9/jobs/46255777, the "of sufficiently low queue-max-size" test failed due to all of the requests actually succeeding despite the lower queue size set.

The particularly challenging aspect of these tests is to generate enough request load on the server that the queue is completely filled while still making these tests more "end-to-end" -- avoiding having to manipulate internal Jetty state and implementation details -- like how many requests are backed up at the different layers - selectors, acceptors, queue, worker threads, ...

I meekly removed this test block in b4da828.  Unit-level tests around setting of the queue size are still in the PR.

camlow325: FWIW, I tried out the following modifications to the previous test and they all passed locally across 20+ iterations.  This approach reduces `max-threads` to the bare minimum needed for the default number of selectors and acceptors and sets `queue-max-size` to the same for the expected failure case.  The number of requests is reduced to be just above this size.

```clj
(deftest queue-max-size-tests
   (testing "queue-max-size"
     (let [response-body  "hello world"
           app            (fn [_]
                            (Thread/sleep 2000)
                            (-> response-body
                            (rr/response)
                            (rr/status 200)
                            (rr/content-type "text/plain")
                            (rr/charset "UTF-8")))
           ;; Boot Jetty with mininum number of threads needed in order to
           ;; reduce the number of requests needed to exhaust the queue for
           ;; the low queue-max-size test case.
           max-threads (jetty-config/calculate-required-threads
                         {:port 0}
                         (ks/num-cpus))
           ;; For the low queue-max-size case, need to generate just enough
           ;; concurrent requests to hit the queue limit and for additional
           ;; requests to be dropped
           num-requests   (+ 2 max-threads)]
           (testing "of default (infinite) size doesn't lead to any request fails"
             (with-test-webserver-and-config
               app port {:max-threads max-threads
                         :shutdown-timeout-seconds 0}
               (let [responses (get-responses num-requests port)]
                 (doseq [response responses]
                   (is (= 200 (:status @response))
                     "Unsuccessful response status for one request")
                   (is (= response-body (:body @response))
                     "Unsuccessful response body for one request")))))
           (testing "of sufficiently low queue-max-size causes some requests to fail"
             (with-test-webserver-and-config
               app port {:max-threads max-threads
                         :queue-max-size max-threads
                         :shutdown-timeout-seconds 0}
               (let [responses (doall (get-responses num-requests port))]
                 (is (some #(try
                   @%
                   false
                   (catch ExecutionException e
                     (instance? ConnectionClosedException
                       (.. e getCause))))
                           responses)
                     "Didn't encounter connection close for any requests")))))))
```

The critical difference this approach has is that the failure test would try to force the queue to overflow because all worker threads would be exhausted fulfilling requests.  With the previous approach, the worker threads would not be exhausted and so having the queue be overflowing would be more timing-dependent.

The above test would also run more slowly in that the reduced number of worker threads to handle the incoming requests would cause request processing on the worker threads to be done more serially.

In any event, I still don't like the race conditions baked into these tests and am not sure we'd want to introduce the fragility into the standard test suite.

fpringvaldsen:
:+1:

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -259,11 +264,27 @@
       (.setHost (:host config)))))
 
 (schema/defn ^:always-validate
+  queue-thread-pool :- QueuedThreadPool
+  [max-threads :- schema/Int
+   queue-max-size :- schema/Int]
+  (let [queue-min-size (if (> default-queue-min-threads queue-max-size)
+                         queue-max-size
+                         default-queue-min-threads)]
```
nwolfe: Isn't this bit of logic equivalent to `(min default-queue-min-threads queue-max-size)`?

camlow325: Yep, and `min` looks better.  Will update that.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -135,11 +133,18 @@
...
 
+(def default-request-idle-timeout
+  1000)
+
 (def default-proxy-idle-timeout
   "The default number of milliseconds to wait before the proxy gives up on the
   upstream server if the :idle-timeout value isn't set in the proxy config."
   60000)
 
+(def default-queue-idle-timeout 60000)
```
nwolfe: If the timeout above didn't have a docstring I wouldn't miss having one on this as much, but it does :smile: I think it would be worthwhile to add a docstring here mentioning what it does and that it's in milliseconds.

camlow325:
In b033efc, added some more docstrings to the constants in jetty9_core.clj and a `with-test-logging` macro for test log suppression to the `calculated-thread-pool-size` test.

nwolfe:
@camlow325 I managed to silence the stacktrace from the test output but wasn't able to get the handful of normal logging statements.

Throw in a `with-test-logging` after the following line: https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj#L673

nwolfe:
:+1: other than stacktrace logging

camlow325:
Good catch, @nwolfe.  Added 8f10239 with one more `with-test-logging` to suppress another error message from the test output.

nwolfe:
:+1:

PR #103, opened 2015-01-24
(SERVER-101) Upgrade jvm-ca dependency

nwolfe:
This updates the jvm-ca dependency to reflect the name change.

This also updates the Trapperkeeper dependency to the newest version.


camlow325:
:+1:

PR #105, opened 2015-02-11
(TK-140) Expose so-linger-seconds setting for webserver connectors

camlow325:
This commit allows the TCP SO_LINGER time used by the sockets underlying
webserver connectors to be configurable via a new webserver setting
called `so-linger-seconds`.  This is an optional setting that, if not
specified, will use the legacy behavior of SO_LINGER being disabled for
webserver sockets.

### DIFF
--- a/doc/jetty-config.md
+++ b/doc/jetty-config.md
@@ -43,7 +43,11 @@ header is being sent in an HTTP Request.
 
 This sets the TCP SO_LINGER time, in seconds, that the webserver uses for
 underlying socket connections.  Values less than 0 result in SO_LINGER
-being disabled.
+being disabled.  Defaults to -1, i.e., "disabled".  For a more detailed
+description of what it means to have SO_LINGER disabled vs. enabled for some
+number of seconds, see http://man7.org/linux/man-pages/man7/socket.7.html.  Note
+that the effect of setting this option may vary depending upon the operating
+system's underlying implementation.
 
 ### `ssl-host`
 
--- a/doc/jetty-config.md
+++ b/doc/jetty-config.md
@@ -39,6 +39,12 @@ that exceeds this value, Jetty will return an HTTP 413 Error response. This
 defaults to 8192 bytes, and only needs to be configured if an exceedingly large
 header is being sent in an HTTP Request.
 
+### `so-linger-seconds`
+
+This sets the TCP SO_LINGER time, in seconds, that the webserver uses for
+underlying socket connections.  Values less than 0 result in SO_LINGER
+being disabled.
+
 ### `ssl-host`
 
 This sets the hostname to listen on for _encrypted_ HTTPS traffic. If not
--- a/project.clj
+++ b/project.clj
@@ -63,7 +63,7 @@
                                   [puppetlabs/trapperkeeper ~tk-version :classifier "test"]
                                   [org.clojure/tools.namespace "0.2.4"]
                                   [org.clojure/java.jmx "0.2.0"]
-                                  [spyscope "0.1.4"]
+                                  [spyscope "0.1.5" :exclusions [clj-time]]
                                   [compojure "1.1.8" :exclusions [ring/ring-core
                                                                   commons-io
                                                                   org.clojure/tools.macro]]]
--- a/project.clj
+++ b/project.clj
@@ -63,7 +63,7 @@
                                   [puppetlabs/trapperkeeper ~tk-version :classifier "test"]
                                   [org.clojure/tools.namespace "0.2.4"]
                                   [org.clojure/java.jmx "0.2.0"]
-                                  [spyscope "0.1.5" :exclusions [clj-time]]
+                                  [spyscope "0.1.4"]
                                   [compojure "1.1.8" :exclusions [ring/ring-core
                                                                   commons-io
                                                                   org.clojure/tools.macro]]]
--- a/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
+++ b/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
@@ -27,7 +27,9 @@
          (update-in [:max-threads] (fnil identity default-max-threads))
          (update-in [:queue-max-size] (fnil identity default-queue-max-size))
          (update-in [:jmx-enable] (fnil ks/parse-bool default-jmx-enable))
-         (update-in [:http :request-header-max-size] (fnil identity default-request-header-size)))
+         (update-in [:http :request-header-max-size] (fnil identity default-request-header-size))
+         (update-in [:http :so-linger-milliseconds] (fnil identity
+                                                          default-so-linger-in-milliseconds)))
      (process-config config)))
 
 (defn expected-https-config?
@@ -41,6 +43,8 @@
            (update-in [:https :protocols] (fnil identity default-protocols))
            (update-in [:https :client-auth] (fnil identity default-client-auth))
            (update-in [:https :request-header-max-size] (fnil identity default-request-header-size))
+           (update-in [:https :so-linger-milliseconds] (fnil identity
+                                                             default-so-linger-in-milliseconds))
            (update-in [:https :ssl-crl-path] identity))
        (-> actual
            (update-in [:https] dissoc :keystore-config)))))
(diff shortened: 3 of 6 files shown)


project.clj
```diff
@@ -62,7 +62,7 @@
                                   [puppetlabs/trapperkeeper ~tk-version :classifier "test"]
                                   [org.clojure/tools.namespace "0.2.4"]
                                   [org.clojure/java.jmx "0.2.0"]
-                                  [spyscope "0.1.4"]
+                                  [spyscope "0.1.5" :exclusions [clj-time]]
```
camlow325: I bumped the spyscope dependency b/c my Intelli-J IDE was complaining about the conflict between the older clj-time dep that the previous spyscope version had vs. the one that other Clojure libraries needed.  This seemed like the cleanest way to overcome the problem.  Also had to bump the clojure dep to 1.6 because the newer spyscope depends upon it.

fpringvaldsen:
:+1: All tests passed locally.

project.clj
cprice404: That's fine, but FYI I've been hitting issues related to clj-time in other projects as well, and I think it's probably also an OK idea to just explicitly specify a version of clj-time for the project.  (I think one of my later J9 PRs does this.)

doc/jetty-config.md
```diff
@@ -39,6 +39,12 @@ that exceeds this value, Jetty will return an HTTP 413 Error response. This
 defaults to 8192 bytes, and only needs to be configured if an exceedingly large
 header is being sent in an HTTP Request.
 
+### `so-linger-seconds`
+
+This sets the TCP SO_LINGER time, in seconds, that the webserver uses for
+underlying socket connections.  Values less than 0 result in SO_LINGER
+being disabled.
```
cprice404: Can you add a bit more verbiage about what "Being disabled" actually means (assuming that doesn't add so much complexity as to make the docs unwieldy)?  And also what the default will be?

cprice404:
+1 other than maybe adding a bit more detail to the docs.

project.clj
camlow325: Fair enough.  I'll go with the approach of just adding a hard clj-time dependency on 0.5.1 and just staying with spyscope 0.1.4 for now.  I'll leave the bump of the clojure dependency to 1.6.0 since it seems like we're going that way for other projects right now anyway.

cprice404: Bumping spyscope is totally fine.

camlow325: Ok, but if I do, I'll have to add an exclusion for it on clj-time to it since spyscope 0.1.5 depends upon a newer clj-time.  'Spose that's okay - not sure though if there's some functionality in spyscope that depends upon functionality in a newer clj-time?

dependencies--

cprice404: what version does spyscope want?  Can't you just add an explicit dep on clj-time too?  But the exclusion isn't really a big deal anyway.

doc/jetty-config.md
camlow325: Yeah, I should definitely document what the default is - which is -1, a.k.a. "disabled".

I had left the description intentionally a little vague since I'm not totally confident that the behaviors I'm seeing at the TCP layer when I use this feature will be consistent across environments -- different OS distributions, etc.  I'll try to beef the description up a bit more, maybe with some caveats about how the actual behaviors may differ per environment.

project.clj
camlow325: spyscope 0.1.4 wants clj-time 0.5.0.  spyscope 0.1.5 wants clj-time 0.7.0.  I think we currently want clj-time 0.5.1 for our TK-related projects.

I can definitely add an explicit dep in the tk-jetty9 project deps for clj-time 0.5.1 and that works with a dependency on spyscope 0.1.5 so long as I add an exclusion for spyscope on clj-time:

```
(defproject puppetlabs/trapperkeeper-webserver-jetty9 "1.1.2-SNAPSHOT"
...
  :dependencies [[org.clojure/clojure "1.6.0"]
                 [org.clojure/tools.logging "0.2.6"]
                 [clj-time "0.5.1"]
...
    :profiles {:dev {
...
      :dependencies
        [spyscope "0.1.5" :exclusions [clj-time]]
```

Are you saying there's a way to have spyscope 0.1.5 explicitly include clj-time 0.7.0 and still have the rest of the project use clj-time 0.5.1?  I was thinking only one clj-time version could ultimately be in use.  Sorry, I'm probably misunderstanding what you're saying...

camlow325: In any event, I didn't have a specific need to bump up spyscope for this PR, so I'm inclined to just stay with the one we're currently using.

cprice404: OK I didn't realize you were saying that it had jumped up all the way to 0.7.0.  I think we do want to stay on 0.5.1 for the rest of the project... so, either explicitly add that and put the exclusion into spyscope like you mentioned, or, leave us on the old spyscope.  Don't care which one.

camlow325:
I added 459fb29 to adjust the project dependencies to include a specific version of clj-time and bump the spyscope dependency back down.  The commit also has some more verbiage in the docs for the `so-linger-seconds` option.

doc/jetty-config.md
```diff
@@ -39,6 +39,16 @@ that exceeds this value, Jetty will return an HTTP 413 Error response. This
...
+### `so-linger-seconds`
+
+This sets the TCP SO_LINGER time, in seconds, that the webserver uses for
+underlying socket connections.  Values less than 0 result in SO_LINGER
+being disabled.  Defaults to -1, i.e., "disabled".  For a more detailed
+description of what it means to have SO_LINGER disabled vs. enabled for some
+number of seconds, see http://man7.org/linux/man-pages/man7/socket.7.html.  Note
+that the effect of setting this option may vary depending upon the operating
+system's underlying implementation.
+
```
jeffmccune: This is a great description of what the configuration setting does, but I'm left wondering what the purpose of the setting is?  Could the description be updated to describe why someone may want to tune this setting?

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
```diff
@@ -93,8 +103,37 @@
...
+                  :request-header-max-size 16192})
+          {:https {:host "foo.local"
+                   :port 8001
+                   :request-header-max-size 16192}}))
+
+    (is (expected-https-config?
+          (merge valid-ssl-pem-config
+                 {:ssl-host "foo.local"
+                  :ssl-port 8001
+                  :so-linger-seconds 22})
```
jeffmccune: I tried setting this to a float, and got an exception.  Since the underlying socket option takes a millisecond value, I think it'd be best to support accepting floating point values so sub-second timeouts can be configured if necessary.

cprice404: if we think subsecond is necessary i'd probably just change the setting to millis

jeffmccune: I'm not sure it's _necessary_, but I have definitely been in situation in the past where I've been frustrated that a piece of software wouldn't let me configure the value I'd calculated as "optimum."

For example, with a load balancer handling thousands of connections we may know that 99.9% of RTT's complete within 150ms, so a 200ms timeout would be more than sufficient to cover "normal" traffic patterns.  However, 1000ms timeout may mean that hundreds (possibly thousands) of connections linger around longer than they need to because the system is handling thousands of connections _per second_, and we're keeping them around 800ms longer than we need to.

Again, I don't have a concrete use case right now, it's just something I've run into working on things at scale previously.  With NGINX, for example, we were easily able to hit 2,500 requests _per second_ with an x1.large in EC2, and many more thousands of open connections at any given time.

jeffmccune:
+1 with my concerns being sub-second configurability and some additional documentation about the purpose of the setting.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
camlow325: I'd agree that it would be better to expose this as milliseconds, however, the underlying API only supports seconds for granularity.  The value that you pass in to the Jetty ServerConnector.setSoLingerTime() - not documented in the method signature - is divided by 1000 as it is passed to the underlying Java Socket:

https://github.com/eclipse/jetty.project/blob/jetty-9.1.0.v20131115/jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java#L355

JavaDocs for the Socket.setSoLinger method include:

```
* Enable/disable SO_LINGER with the specified linger time in seconds.
```

It makes sense that Java constrains this to seconds in that the underlying BSD SO_LINGER option is also expressed in seconds.  From the SO_LINGER documentation in http://man7.org/linux/man-pages/man7/socket.7.html:

```c
struct linger {
  int l_onoff;    /* linger active */
  int l_linger;   /* how many seconds to linger for */
};
```

jeffmccune: Ah, well in that case it's a moot point and nevermind.  Sorry, I even read
the BSD man page and it didn't sink in since the JVM layer exposed the
setting as milliseconds.

doc/jetty-config.md
camlow325: This may only be used in rare circumstances.  I hesitate to say anything more sweeping like "if you have a load balancer routing traffic to your web server and your webserver is leaking file descriptors, this could clean them up".  What if I added something like this?

*You may consider enabling SO_LINGER if non-well behaving clients prevent the webserver from gracefully closing out server sockets properly.  This should be considered more of a "last resort", however.  In such scenarios, it would be better to find a way for clients to terminate socket connections more gracefully.*

@cprice404 - what do you think?

nwolfe:
:+1:

PR #106, opened 2015-02-12
(TK-144) Add support for configuring Jetty via script

cprice404:
A problem that we have had a few times with Jetty is that, in order to try to
make it easy for our users to manage the application without having to be Jetty
experts, we run Jetty embedded and only surface a subset of its configuration
options that we believe the users might care about.

This provides a pretty nice UX in cases where we've plumbed the appropriate
settings through into our config before shipping a build, but, it can be
problematic when we are trying to debug a customer issue and we discover a
potentially-relevant Jetty setting that we haven't wired through to our
configuration files... it means that we can't toggle the Jetty config setting
without shipping them a new build.

After discovering how logback supports some basic Java scripting in their
EvaluatorFilters (via a library called Janino), it occurred to me that we might
be able to support something similar for our Jetty configuration.

This commit does that.  It creates a setting called 'post-config-script', which
can be used for advanced use cases to provide some code to manipulate a Jetty
`Server` object programatically. This should theoretically allow us to cobble
together some sample configuration to ship to users who already have a build
installed that can modify basically any setting that is available in Jetty.

Use cases for this setting would hopefully be temporary, and when we find new
Jetty settings that we determine would be valuable to users, we can just plumb
them through to our normal configuration for the next release, but this would
give users a workaround that doesn't require us to ship them a whole new build
in the interim.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
@@ -546,6 +546,8 @@
                                maybe-logged)]
       (.setHandler s statistics-handler)
       (.setStopTimeout s (or shutdown-timeout default-graceful-stop-timeout))
+      (when-let [script (:post-config-script options)]
+        (config/execute-post-config-script! s script))
       (assoc webserver-context :server s))))
 
 (schema/defn ^:always-validate start-webserver! :- ServerContext
--- a/doc/jetty-config.md
+++ b/doc/jetty-config.md
@@ -267,6 +267,31 @@ the response, and the size in bytes of the response.
 Optional. This is an integer representing the desired graceful stop timeout in seconds.
 Defaults to 60 seconds.
 
+### `post-config-script`
+
+Optional.  This setting is for advanced use cases only, and is intended for
+debugging purposes.  You can use it to modify low-level Jetty settings that
+are not directly exposed in our normal configuration options.  In most cases,
+if you find yourself using this, it is an indicator that we need to expose
+additional settings directly in our main configuration (so please let us know!).
+Also, the implementation details of this setting may change between releases.
+
+If you do need to use this, you can set the value to a String containing some
+Java code that should be executed against the Jetty `Server` object.  This object
+will be injected into the scope of your code in a variable named `server`.
+
+Here is a pathological example that shows how you could change the port that
+your server listens on (which you could achieve in a much simpler fashion by
+using the existing `port` setting; this example is only for the purposes of
+illustration):
+
+    post-config-script: "import org.eclipse.jetty.server.ServerConnector;
+                         ServerConnector c = (ServerConnector)(server.getConnectors()[0]);
+                         c.setPort(10000);"
+
+For more info on the Jetty `Server` object model, see the
+[Jetty Javadocs](http://download.eclipse.org/jetty/stable-9/apidocs/org/eclipse/jetty/server/Server.html).
+
 ## Configuring multiple webservers on isolated ports
 
 It is possible to configure multiple webservers on isolated ports within a single Jetty9
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(def tk-version "1.0.2-SNAPSHOT")
+(def tk-version "1.1.0")
 (def ks-version "1.0.0")
 
 (defproject puppetlabs/trapperkeeper-webserver-jetty9 "1.1.2-SNAPSHOT"
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(def tk-version "1.0.1")
+(def tk-version "1.0.2-SNAPSHOT")
 (def ks-version "1.0.0")
 
 (defproject puppetlabs/trapperkeeper-webserver-jetty9 "1.1.2-SNAPSHOT"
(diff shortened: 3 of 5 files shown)


fpringvaldsen:
:see_no_evil: This seems like it's going to be really useful! It's looking good so far, @cprice404 I think it just needs some docs as you suggested and a new release of TK to get rid of the SNAPSHOT dependency.

KevinCorcoran:
I'm :+1: on this, in general.  I took a quick look at the code and don't see any major problems.

However, I'm interested if you considered any alternative solutions to this problem.  There are already a lot of ways to configure Jetty ... in particular, it seems like we might be able to solve this problem by adding a hook to support Jetty's normal `web.xml` configuration.  [This example](https://wiki.eclipse.org/Jetty/Tutorial/Embedding_Jetty#Configuring_a_File_Server_with_XML) seems to show a path forward that we might be able to follow.  Everybody hates XML, but it *is* baked-into Jetty, so perhaps there's advantages to that in this case.

I don't have a strong opinion on `web.xml` vs. Janino, but curious what you think about that option.

cprice404:
Mostly I don't want to introduce any new config files / syntaxes that we have to support.  This approach seems like it gives us the flexibility we'd need for extremely advanced use cases w/o having to open up the whole can of worms.  We also already have a dependency on Janino now via logback.  I'm open to further discussion, though.

KevinCorcoran:
After thinking about this a bit more, I think that Janino fits-in better with both the primary use-case for this feature (debugging difficult issues on a deployed application) and the way in which this library is set-up (exposing Jetty configuration bit-by-bit, as needed).

The best reason to do XML-based configuration instead would be that it seems a bit more useful in general, as a feature of this library, while a bit less-convenient in other cases.

It might be interesting to see what others think - @camlow325? - but I'm fine with the choice of Janino.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -546,6 +546,8 @@
                                maybe-logged)]
       (.setHandler s statistics-handler)
       (.setStopTimeout s (or shutdown-timeout default-graceful-stop-timeout))
+      (when-let [script (:post-config-script options)]
```
camlow325: Would it be better to invoke the script here or right before `.start` is called on the server in `start-webserver!`?  Even though that doesn't appear to be the case right now, just thinking that if we ever had some logic in between the end of `create-webserver` and the actual starting of the server, it would be good for the script to be executed at the last possible point that a configuration override would need to be made.

cprice404: Fair question.  I think my logic was that the code would be more maintainable in the long run if all of the config stuff was all consolidated in a single place in the code... so I think that my gut is that if someone was to submit a future PR that did config stuff after this function was called, we'd probably want to encourage them to move it into this function instead?  I could be persuaded, though.

camlow325: Yeah, I think best practice would be to do all of the configuration here.  Just wasn't sure if I trust us to always follow that practice.  I'd probably still lean toward putting it in `start-webserver!`, but I understand the argument for leaving it here, too.  I'll defer to your judgement on this.

cprice404: Well, in that case it sounds like it comes down to "should I break from our 'best practice' now, in order to try be defensive about someone else breaking from our 'best practice' in the future?"  To which I'd lean "no".  :)

So I'd rather leave it here unless you try harder to convince me :)

camlow325:
To @KevinCorcoran's comment, I prefer this approach with Janino over the static XML approach as well.

One of the upfront things that would concern me with the XML configuration approach is that I think we want to have this hook invoked after the programmatic webserver configuration that the tk-jetty9 service already does is complete.  Jetty's XML configuration methods seem to assume that they will produce a Server rather than being applied over the top of an existing configuration.

Also, the programmatic approach gives you the ability to read the default configuration of the Server and dynamically reconfigure the Server based on what is found.  This seems like a key capability that Janino would open up that I don't see how the XML approach would allow for.

camlow325:
This probably falls more in the realm of an example worth covering in documentation to be created for this feature, so maybe not relevant to changes to include in the PR itself...

One of the use cases that I envision could be really valuable for this functionality is to be able to log interesting parts of the Server configuration out through standard JVM log facilities.  I suppose you'd be able to create an Slf4j logger in the post-config-script snippet and make calls to it in order to have your custom logged messages appear along with the rest of the logging that services in the JVM stack produce?

camlow325:
:+1: pending docs and tk dependency update.  Probably needs to be rebased on the latest tk-jetty9 changes in order for the SSLv3 unit tests to pass in Travis properly.

This is very cool stuff!

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
camlow325: No, I'm good.  :)

cprice404:
@camlow325 added a commit with docs and a bump to the TK version.

fpringvaldsen:
:+1: On the new docs.

camlow325:
:+1:

PR #107, opened 2015-02-13
(TK-141) Upgrade Jetty dependency to 9.2.7.v20150116

camlow325:
This commit upgrades this service's Jetty dependency to 9.2.7.v20150116,
along with a few corresponding changes needed in the production and test
code for compatibility with the new version.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
@@ -324,15 +324,15 @@
 
 (defn selectors-count
   "The number of selector threads that should be allocated per connector per
-  core.  This algorithm duplicates the default that Jetty 9.2.7.v20150116 uses.
-  See: https://github.com/eclipse/jetty.project/blob/jetty-9.2.7.v20150116/jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java#L229"
+  core.  This algorithm duplicates the default that Jetty 9.2.8.v20150217 uses.
+  See: https://github.com/eclipse/jetty.project/blob/jetty-9.2.8.v20150217/jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java#L229"
   [num-cpus]
   (max 1 (min 4 (int (/ num-cpus 2)))))
 
 (defn acceptors-count
   "The number of acceptor threads that should be allocated per connector per
-  core.  This algorithm duplicates the default that Jetty 9.2.7.v20150116 uses.
-  See: https://github.com/eclipse/jetty.project/blob/jetty-9.2.7.v20150116/jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java#L190"
+  core.  This algorithm duplicates the default that Jetty 9.2.8.v20150217 uses.
+  See: https://github.com/eclipse/jetty.project/blob/jetty-9.2.8.v20150217/jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java#L190"
   [num-cpus]
   (max 1 (min 4 (int (/ num-cpus 8)))))
 
@@ -354,8 +354,8 @@
   calculate-required-threads :- schema/Int
   "Calculate the number threads needed to operate based on the number of cores
   available and the number of connectors present.  This algorithm duplicates
-  the default that Jetty 9.2.7.v20150116 uses.  See:
-  https://github.com/eclipse/jetty.project/blob/jetty-9.2.7.v20150116/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java#L334-L350"
+  the default that Jetty 9.2.8.v20150217 uses.  See:
+  https://github.com/eclipse/jetty.project/blob/jetty-9.2.8.v20150217/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java#L334-L350"
   [config   :- WebserverRawConfig
    num-cpus :- schema/Int]
   (+ 1 (* (connector-count config)
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
@@ -323,14 +323,18 @@
              "Error: More than one default server specified in configuration"))))
 
 (defn selectors-count
-  "The number of selector threads that should be allocated per connector per core."
+  "The number of selector threads that should be allocated per connector per
+  core.  This algorithm duplicates the default that Jetty 9.2.7.v20150116 uses.
+  See: https://github.com/eclipse/jetty.project/blob/jetty-9.2.7.v20150116/jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java#L229"
   [num-cpus]
-  num-cpus)
+  (max 1 (min 4 (int (/ num-cpus 2)))))
 
 (defn acceptors-count
-  "The number of acceptor threads that should be allocated per connector per core."
+  "The number of acceptor threads that should be allocated per connector per
+  core.  This algorithm duplicates the default that Jetty 9.2.7.v20150116 uses.
+  See: https://github.com/eclipse/jetty.project/blob/jetty-9.2.7.v20150116/jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java#L190"
   [num-cpus]
-  (max 1 (int (/ num-cpus 2))))
+  (max 1 (min 4 (int (/ num-cpus 8)))))
 
 (defn threads-per-connector
   "The total number of threads needed per attached connector."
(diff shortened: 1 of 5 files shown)


src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -363,12 +369,11 @@
...
-  (or (:max-threads config)
-      (let [calc-max-threads (max (calculate-required-threads config num-cpus)
-                                  default-max-threads)]
-        (log/warn "Thread pool size not configured so using a size of "
-                  calc-max-threads)
-        calc-max-threads)))
+  (let [max-threads (or (:max-threads config)
+                        (max (calculate-required-threads config num-cpus)
+                             default-max-threads))]
+    (log/debugf "Using webserver thread pool size of %d" max-threads)
```
camlow325: This came out of some informal discussions about the usefulness of the warn message.  Thought it might be better to have a way to see the number of threads that the server is going to use regardless of the configuration but only do it at a debug level since it's probably not as interesting for standard usage.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -393,6 +398,7 @@
   (let [handler (RequestLogHandler.)
         logger (RequestLogImpl.)]
     (.setFileName logger (:access-log-config config))
+    (.setQuiet logger true)
```
camlow325: This gets us to 100% compliance for Wolfe's Principle 1 on keeping logging silent for test runs - for now.  Without this change, a few verbose logback "-INFO" messages are written out when the logger is initialized, like these:

```
11:55:59,250 |-INFO in LogbackRequestLog - Will use configuration file [./dev-resources/puppetlabs/trapperkeeper/services/webserver/request-logging.xml]
11:55:59,254 |-INFO in ch.qos.logback.access.joran.action.ConfigurationAction - debug attribute not set
11:55:59,254 |-INFO in ch.qos.logback.core.joran.action.AppenderAction - About to instantiate appender of type [appender.TestListAppender]
11:55:59,254 |-INFO in ch.qos.logback.core.joran.action.AppenderAction - Naming appender as [LIST]
```

These don't seem particularly useful and probably not worth making configurable for now.

camlow325:
This PR should come in only after other PRs already in flight - specifically #105 and #106.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_proxy_test.clj
```diff
@@ -133,6 +164,24 @@
         (is (= (:status response) 200))
         (is (= (:body response) "Hello, World!")))))
 
+  (testing "basic proxy support when proxy handler registered after server start"
```
fpringvaldsen: Shouldn't this be before server start?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -323,14 +323,18 @@
              "Error: More than one default server specified in configuration"))))
 
 (defn selectors-count
-  "The number of selector threads that should be allocated per connector per core."
+  "The number of selector threads that should be allocated per connector per
+  core.  This algorithm duplicates the default that Jetty 9.2.7.v20150116 uses.
+  See: https://github.com/eclipse/jetty.project/blob/jetty-9.2.7.v20150116/jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java#L229"
   [num-cpus]
-  num-cpus)
+  (max 1 (min 4 (int (/ num-cpus 2)))))
```
fpringvaldsen: Did the Jetty defaults change since 9.1.0, or were we just not using the Jetty defaults for these?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -323,14 +323,18 @@
...
 
 (defn acceptors-count
-  "The number of acceptor threads that should be allocated per connector per core."
+  "The number of acceptor threads that should be allocated per connector per
+  core.  This algorithm duplicates the default that Jetty 9.2.7.v20150116 uses.
+  See: https://github.com/eclipse/jetty.project/blob/jetty-9.2.7.v20150116/jetty-server/src/main/java/org/eclipse/jetty/server/AbstractConnector.java#L190"
   [num-cpus]
-  (max 1 (int (/ num-cpus 2))))
+  (max 1 (min 4 (int (/ num-cpus 8)))))
```
fpringvaldsen: Same question here as above.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
fpringvaldsen: Nice! I couldn't figure out how to suppress this logging in the Access Logging tests when I added them. Glad to see these log messages are going away.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_proxy_test.clj
```diff
@@ -68,28 +69,58 @@
...
+          "/goodbye"))
+      (if ~register-proxy-route-before-server-start?
+        (let [proxy-service# (proxy-service ~proxy-config
+                                            ~proxy-opts
+                                            ~proxy-path)]
+          (with-app-with-config proxy-app#
+            [jetty9-service proxy-service#]
+            {:webserver ~proxy}
+            ~@body))
+          (with-app-with-config proxy-app#
```
fpringvaldsen: This is super nitpicky so feel free to ignore it, but I was initially confused by this indentation and thought this was part of the `let` block above.

fpringvaldsen:
Looks good! I'm glad to see we've finally figured out the issues around upgrading Jetty. I had a couple minor comments/questions above, other than that I'm :+1:.

camlow325:
I had to rebase this when #108 came in so I just took care of that and the couple of other problems that @fpringvaldsen found -  the before/after mixup in the test description and alignment on one form - in bb824d1.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
camlow325: Yeah, the defaults that Jetty uses internally did change between 9.1.0 and 9.2.7.  Our current implementation on tk-jetty9 master matches the Jetty 9.1.0 defaults.  This PR would make the tk-jetty9 implementation mirror the Jetty 9.2.7 implementation for these defaults.  Arguably, it would be better instead to keep the defaults the same in the tk-jetty9 service for compatibility.  I talked with @cprice404 about this a bit and we were thinking it might be better to just follow core Jetty's lead since the maintainers undoubtedly have more insight than we do on how to tune these.

It may make sense sometime down the road to provide a way for selectors and acceptors to be configured externally, i.e., through the webserver configuration.  At that point, honoring the values set in the configuration file across releases would be a must but, in the absence of that, these values are more of an internal configuration detail and it may be best for now to follow core Jetty's lead for the auto-tuning.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
camlow325: Same response here as for the `selectors-count` above.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
fpringvaldsen: Ah, I see. I'm on board with this change, then.

fpringvaldsen:
:+1:

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
KevinCorcoran: :+1: the old warning is making me twitchy.

I'd be okay with this being an INFO instead of DEBUG, but, either way is better.

nwolfe:
@camlow325 Is this still blocked on #106 coming in first?

camlow325:
@nwolfe Yeah, I think we still should wait for #106 to come in first.  Also, I'm debating rebasing this up to the new Jetty which just released, 9.2.8, as it appears to have some fixes related to stuff that we use like ProxyServlet and GzipHandler.  @cprice404 - WDYT?

camlow325:
In 989b776, I updated the Jetty dependency to 9.2.8.v20150217 so that we could incorporate the ProxyServlet and GzipHandler fixes.

cprice404:
:see_no_evil: I reviewed this in a little detail except for the proxy test changes, so I'm basically +1 after #106 comes in.

KevinCorcoran:
Would anybody be willing to merge this, or object to me doing so?  I'd like to upgrade our jetty dependency to bring us into compliance with Wolfe's First Principal.

PR #109, opened 2015-02-27
(TK-133) Allow comma separated strings in ssl-protocols, cipher-suites

mullr:
- Also turns a higher-level use of schema/either into schema/conditional, for better error messages
- The docs around the defaults for these options (https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/doc/jetty-config.md#cipher-suites) appear to be wrong, but I may be missing something subtle. Let me know if you want me to update them.

### DIFF
--- a/doc/jetty-config.md
+++ b/doc/jetty-config.md
@@ -150,25 +150,38 @@ This sets the passphrase to use for unlocking the truststore file.
 
 ### `cipher-suites`
 
-Optional. A comma-separated list of cryptographic ciphers to allow for incoming
-SSL connections. Valid names are listed in the
+Optional. The cryptographic ciphers to allow for incoming SSL
+connections. This may be formatted either as a list (in a HOCON
+configuration file) or a comma-separated string. Valid names are
+listed in the
 [official JDK cryptographic providers documentation](http://docs.oracle.com/javase/7/docs/technotes/guides/security/SunProviders.html#SupportedCipherSuites);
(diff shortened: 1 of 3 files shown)


KevinCorcoran:
This looks good to me, aside from one nitpick.  @camlow325 and/or @fpringvaldsen should probably review as well.  Thanks for the fix!

mullr:
@KevinCorcoran Fixed the formatting and took out the always-validate. Let me know if there's anything else!

KevinCorcoran:
:+1:

camlow325:
Yeah, I think you're right that the `cipher-suites` defaults are documented incorrectly.  I think the defaults always come from the vector at https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/trapperkeeper-webserver-jetty9-1.1.1/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj#L16-L23.  I think it would be better to just explicitly document those and remove all of the JDK-version specific and non-DHE language.

@cprice404 - does that sound right to you?

camlow325:
:+1: on the implementation.  I think correcting the docs on `cipher-suites` would be good to do as part of this PR as well, while we're still thinking about it.

mullr:
Ok, I updated the docs for both `cipher-suites` and `ssl-protocols`, since that had the same issue.

cprice404:
sure

camlow325:
:+1:

PR #110, opened 2015-03-01
(TK-151) Add idle timeout setting

cprice404:
This PR adds support for a new setting, `idle-timeout-milliseconds`, which can be used to control how long Jetty will allow a client connection to remain open with no activity.

### DIFF
--- a/doc/jetty-config.md
+++ b/doc/jetty-config.md
@@ -55,6 +55,8 @@ This optional setting can be used to control how long Jetty will allow a
 connection to be held open by a client without any activity on the socket.  If
 a connection is idle for longer than this value, Jetty will forcefully close
 it from the server side.  Jetty's default value for this setting is 30 seconds.
+Note that Jetty will not automatically close the connection if the idle timeout
+is reached while Jetty is still actively processing a client request.
 
 ### `ssl-host`
 
--- a/doc/jetty-config.md
+++ b/doc/jetty-config.md
@@ -49,6 +49,13 @@ number of seconds, see http://man7.org/linux/man-pages/man7/socket.7.html.  Note
 that the effect of setting this option may vary depending upon the operating
 system's underlying implementation.
 
+### `idle-timeout-milliseconds`
+
+This optional setting can be used to control how long Jetty will allow a
+connection to be held open by a client without any activity on the socket.  If
+a connection is idle for longer than this value, Jetty will forcefully close
+it from the server side.  Jetty's default value for this setting is 30 seconds.
+
 ### `ssl-host`
 
 This sets the hostname to listen on for _encrypted_ HTTPS traffic. If not
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
@@ -58,6 +58,7 @@
    (schema/optional-key :queue-max-size)             schema/Int
    (schema/optional-key :request-header-max-size)    schema/Int
    (schema/optional-key :so-linger-seconds)          schema/Int
+   (schema/optional-key :idle-timeout-milliseconds)  schema/Int
    (schema/optional-key :ssl-port)                   schema/Int
    (schema/optional-key :ssl-host)                   schema/Str
    (schema/optional-key :ssl-key)                    schema/Str
@@ -120,7 +121,8 @@
 
 (def WebserverConnectorCommon
   {:request-header-max-size schema/Int
-   :so-linger-milliseconds  schema/Int})
+   :so-linger-milliseconds  schema/Int
+   :idle-timeout-milliseconds (schema/maybe schema/Int)})
 
 (def WebserverConnector
   (merge WebserverConnectorCommon
@@ -325,7 +327,8 @@
   [config :- WebserverRawConfig]
   {:request-header-max-size   (or (:request-header-max-size config)
                                   default-request-header-size)
-   :so-linger-milliseconds    (so-linger-in-milliseconds config)})
+   :so-linger-milliseconds    (so-linger-in-milliseconds config)
+   :idle-timeout-milliseconds (:idle-timeout-milliseconds config)})
 
 (schema/defn ^:always-validate
   maybe-get-http-connector :- (schema/maybe WebserverConnector)
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
@@ -118,11 +118,14 @@
 (def WebserverSslClientAuth
   (schema/enum :need :want :none))
 
-(def WebserverConnector
-  {:host                    schema/Str
-   :port                    schema/Int
-   :request-header-max-size schema/Int
+(def WebserverConnectorCommon
+  {:request-header-max-size schema/Int
    :so-linger-milliseconds  schema/Int})
+
+(def WebserverConnector
+  (merge WebserverConnectorCommon
+         {:host schema/Str
+          :port schema/Int}))
 
 (def WebserverSslContextFactory
   {:keystore-config                    WebserverSslKeystoreConfig
@@ -318,13 +321,19 @@
     default-so-linger-in-milliseconds))
 
 (schema/defn ^:always-validate
+  common-connector-config :- WebserverConnectorCommon
+  [config :- WebserverRawConfig]
+  {:request-header-max-size   (or (:request-header-max-size config)
+                                  default-request-header-size)
+   :so-linger-milliseconds    (so-linger-in-milliseconds config)})
+
+(schema/defn ^:always-validate
   maybe-get-http-connector :- (schema/maybe WebserverConnector)
   [config :- WebserverRawConfig]
   (if (contains-http-connector? config)
-    {:host         (or (:host config) default-host)
-     :port         (or (:port config) default-http-port)
-     :request-header-max-size (or (:request-header-max-size config) default-request-header-size)
-     :so-linger-milliseconds (so-linger-in-milliseconds config)}))
+    (merge (common-connector-config config)
+           {:host (or (:host config) default-host)
+            :port (or (:port config) default-http-port)})))
 
 (schema/defn ^:always-validate
   contains-https-connector? :- schema/Bool
@@ -335,15 +344,14 @@
   maybe-get-https-connector :- (schema/maybe WebserverSslConnector)
   [config :- WebserverRawConfig]
   (if (contains-https-connector? config)
-    {:host (or (:ssl-host config) default-host)
-     :port (or (:ssl-port config) default-https-port)
-     :request-header-max-size (or (:request-header-max-size config) default-request-header-size)
-     :so-linger-milliseconds (so-linger-in-milliseconds config)
-     :keystore-config (get-keystore-config! config)
-     :cipher-suites (get-cipher-suites-config config)
-     :protocols (get-ssl-protocols-config config)
-     :client-auth (get-client-auth! config)
-     :ssl-crl-path (get-ssl-crl-path! config)}))
+    (merge (common-connector-config config)
+           {:host                    (or (:ssl-host config) default-host)
+            :port                    (or (:ssl-port config) default-https-port)
+            :keystore-config         (get-keystore-config! config)
+            :cipher-suites           (get-cipher-suites-config config)
+            :protocols               (get-ssl-protocols-config config)
+            :client-auth             (get-client-auth! config)
+            :ssl-crl-path            (get-ssl-crl-path! config)})))
 
 (schema/defn ^:always-validate
   maybe-add-http-connector :- {(schema/optional-key :http) WebserverConnector
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
@@ -186,8 +186,6 @@
       (log/info (str "webserver config overridden for key '" (name key) "'")))
     (merge options overrides)))
 
-
-
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; SSL Context Functions
 
@@ -260,16 +258,19 @@
    config :- (merge config/WebserverConnector
                     {schema/Keyword schema/Any})
    ssl-ctxt-factory :- (schema/maybe SslContextFactory)]
-  (let [request-size (:request-header-max-size config)]
-    (doto (ServerConnector.
-            server
-            nil nil nil
-            (config/acceptors-count (ks/num-cpus))
-            (config/selectors-count (ks/num-cpus))
-            (connection-factories request-size ssl-ctxt-factory))
-      (.setPort (:port config))
-      (.setHost (:host config))
-      (.setSoLingerTime (:so-linger-milliseconds config)))))
+  (let [request-size (:request-header-max-size config)
+        connector   (doto (ServerConnector.
+                            server
+                            nil nil nil
+                            (config/acceptors-count (ks/num-cpus))
+                            (config/selectors-count (ks/num-cpus))
+                            (connection-factories request-size ssl-ctxt-factory))
+                      (.setPort (:port config))
+                      (.setHost (:host config))
+                      (.setSoLingerTime (:so-linger-milliseconds config)))]
+    (when-let [idle-timeout (:idle-timeout-milliseconds config)]
+      (.setIdleTimeout connector idle-timeout))
+    connector))
 
 (schema/defn ^:always-validate
   ssl-connector  :- ServerConnector
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
@@ -242,13 +242,34 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Jetty Server / Connector Functions
 
-(defn- connection-factory
-  [request-header-size]
+(defn- connection-factories
+  [request-header-size ssl-ctxt-factory]
   (let [http-config (doto (HttpConfiguration.)
                       (.setSendDateHeader true)
-                      (.setRequestHeaderSize request-header-size))]
-    (into-array ConnectionFactory
-                [(HttpConnectionFactory. http-config)])))
+                      (.setRequestHeaderSize request-header-size))
+        factories   (into-array ConnectionFactory
+                                [(HttpConnectionFactory. http-config)])]
+    (if ssl-ctxt-factory
+      (AbstractConnectionFactory/getFactories
+        ssl-ctxt-factory factories)
+      factories)))
+
+(schema/defn ^:always-validate
+  connector* :- ServerConnector
+  [server :- Server
+   config :- (merge config/WebserverConnector
+                    {schema/Keyword schema/Any})
+   ssl-ctxt-factory :- (schema/maybe SslContextFactory)]
+  (let [request-size (:request-header-max-size config)]
+    (doto (ServerConnector.
+            server
+            nil nil nil
+            (config/acceptors-count (ks/num-cpus))
+            (config/selectors-count (ks/num-cpus))
+            (connection-factories request-size ssl-ctxt-factory))
+      (.setPort (:port config))
+      (.setHost (:host config))
+      (.setSoLingerTime (:so-linger-milliseconds config)))))
 
 (schema/defn ^:always-validate
   ssl-connector  :- ServerConnector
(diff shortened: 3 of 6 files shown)


doc/jetty-config.md
```diff
@@ -49,6 +49,13 @@ number of seconds, see http://man7.org/linux/man-pages/man7/socket.7.html.  Note
 that the effect of setting this option may vary depending upon the operating
 system's underlying implementation.
 
+### `idle-timeout-milliseconds`
+
+This optional setting can be used to control how long Jetty will allow a
+connection to be held open by a client without any activity on the socket.  If
+a connection is idle for longer than this value, Jetty will forcefully close
+it from the server side.  Jetty's default value for this setting is 30 seconds.
```
camlow325: I might include something about the following paragraph in the description, but I'll defer to your judgement...

From what I've seen, Jetty will only forcefully close the socket per the idle timeout if a client request is not being actively handled by the server.  For example, if the client had issued a full HTTP request, the server is still in the process of preparing a response, and no subsequent activity had taken place on the socket up to the configured idle timeout, Jetty will not forcefully close the connection from the server side.  Instead, Jetty will keep the socket alive from its side through the delivery of the response.  Once the response has been delivered, if no subsequent activity on the socket occurs up through the configured idle timeout again, Jetty would close the socket.

Jetty also uses the idle timeout as part of its internal algorithm for scaling up / down worker threads.  If one of its worker threads has been idle for a period up to the idle timeout, Jetty will let that thread die - so long as Jetty's worker thread pool has at least the configured minimum number of threads still available to it.  I'm thinking that the threading behavior may be more of an internal detail that's not worth trying to document here.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -318,13 +323,20 @@
     default-so-linger-in-milliseconds))
 
 (schema/defn ^:always-validate
+  common-connector-config :- WebserverConnectorCommon
+  [config :- WebserverRawConfig]
+  {:request-header-max-size   (or (:request-header-max-size config)
+                                  default-request-header-size)
+   :so-linger-milliseconds    (so-linger-in-milliseconds config)
+   :idle-timeout-milliseconds (:idle-timeout-milliseconds config)})
```
camlow325: I understand that Jetty's default for idle timeout is currently 30 seconds.  Elsewhere in the tk-jetty9 service, though, I think we've more defensively coded in our own local default that we pass down to the Jetty configuration - rather than just relying upon default Jetty behavior.  I like this approach since it protects us from Jetty changing the default in a newer version without us noticing.  What would you think of coding the 30 second default in as a constant in tk-jetty9 and having tk-jetty9 always set a value for the connector's idle timeout?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -318,13 +323,20 @@
     default-so-linger-in-milliseconds))
 
 (schema/defn ^:always-validate
+  common-connector-config :- WebserverConnectorCommon
```
camlow325: Good call on factoring this out.  The settings repetition between the http and https connectors was starting to get ugly.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -242,13 +240,37 @@
...
-                [(HttpConnectionFactory. http-config)])))
+                      (.setRequestHeaderSize request-header-size))
+        factories   (into-array ConnectionFactory
+                                [(HttpConnectionFactory. http-config)])]
+    (if ssl-ctxt-factory
+      (AbstractConnectionFactory/getFactories
+        ssl-ctxt-factory factories)
+      factories)))
+
+(schema/defn ^:always-validate
```
camlow325: Nice!

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
```diff
@@ -301,3 +311,53 @@
...
+          config
+          (let [s (tk-app/get-service app :WebserverService)
+                add-ring-handler (partial add-ring-handler s)]
+            (add-ring-handler ring-handler path)
+
+            (testing "Verify that server doesn't close socket before idle timeout"
+              ;; if we set the client socket timeout lower than the server
+              ;; socket timeout, we should get a timeout exception from the
+              ;; client side while attempting to read from the socket.
+              (is (thrown-with-msg? SocketTimeoutException #"Read timed out"
```
camlow325: I've run this test 50 or so times locally without seeing it fail.  It does seem like it could fail given that the server might start its idle timeout counter a bit sooner after writing the last byte of the response to the client than the client would after consuming the last byte of the response received back from the server.  I'm okay leaving this in for now, but we may need to revisit this if the test starts failing in CI.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
```diff
@@ -301,3 +311,53 @@
...
+                                    (read-response 250))))
+            (testing "Verify that server closes the socket after idle timeout"
+              ;; if we set the client socket timeout higher than the server,
+              ;; then the server should close the socket after its timeout,
+              ;; which will cause our read to stop blocking and allow us to
+              ;; validate the contents of the data we read from the socket.
+              (let [resp (read-response 750)]
```
camlow325: I also ran this one 50 or so times locally without seeing it fail.  It does seem like the precision around the timeout is difficult to measure effectively with this test.  For example, if the server happened to actually use more like a 30 second idle timeout instead of the 500 millisecond one the test configures, I think this test would still pass.  It may make sense to add some extra validation to this test to see that the total time spent in the `read-response` call doesn't take more than some amount of time - maybe a couple of seconds - although, admittedly, the extra validation might make the test more fragile.

doc/jetty-config.md
camlow325: Scratch my last comment on the idle timeout as it pertains to worker threads being idle.  That is actually controlled by the idle timeout passed into the `QueuedThreadPool` constructor:

https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/trapperkeeper-webserver-jetty9-1.1.1/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj#L280

We're hard-coding that to 60 seconds right now, following what Jetty does if you call one of the other forms of the constructor that don't allow you to specify the idle timeout:

https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/trapperkeeper-webserver-jetty9-1.1.1/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj#L145-L149

So, there doesn't seem to be anything relevant here to discuss for this setting.

camlow325:
For the last couple of new server settings I plumbed through - so-linger-seconds and queue-max-size - I'd started adding some tests under `create-server-tests` in "jetty9_core_test.clj".  These tests basically just query back on the Jetty Server instance that tk-jetty9 created to see that the Server is configured with the values we were expecting it to be configured with for specific settings.  These aren't as end-to-end as are the tests that involve making HTTP client requests to a running server, but they do allow for more precise value testing.

For example, I could set a "seconds" based setting to "421034" and still have the test ensure that that exact value would be used by Jetty.  Also, test would run in a very short amount of time, making it more practical to be run during local development.

It may be worthwhile to do similar tests for the new `idle-timeout-milliseconds` setting.  This could help mitigate with the concern I had about the one `idle-timeout-milliseconds` setting test which might pass even if the effective timeout used by Jetty was very different from the one we had configured.

camlow325:
:+1: with just the questions on:

* possibly adding more info to the docs about the behavior of the timeout setting when a request is still active.

* possibly coding the default `idle-timeout` setting into tk-jetty9 itself rather than just relying upon default Jetty behavior when the setting isn't specified in the Trapperkeeper webserver configuration.

* possibly adding more validation to the second `idle-timeout-milliseconds` test to see that the total request time isn't taking too long.

* possibly adding an `idle-timeout` test or two which queries back on the Jetty ServerConnector to see that it was configured with the idle timeout value(s) that we were expecting,

doc/jetty-config.md
cprice404: So, you're saying that the docs are OK as-is?

camlow325: The part about Jetty not forcefully closing the connection if the socket is idle while a request is still in process is still relevant.  I think this might be good to cover in the docs here but wanted to get your opinion on if you thought this was worthwhile.

cprice404: I'm somewhere between "is this something that a user will really want/need to know the details of" and "is there even a way that I can concisely describe this?"

I'm willing to patch the docs if you think it's important and can give me a description that makes sense and doesn't double the length of the current docs :)

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
cprice404: We should have a more broad discussion about this with the team; let's figure out how to get that scheduled.

I'm firmly in the "no" camp here; I like the idea of making things configurable but I'm pretty much terrified of making assumptions that we will have more insight into what the sane default values should be than the upstream authors would.

Think about this in light of the tuning work we're about to do in Puppet Server.  If someone had written their own wrapper around our installer that resulted in a hard-coded override of our internal defaults, to set them to what we currently set them to, and then we roll out Puppet Server 1.1 with a bunch of improvements to those defaults based on our own domain expertise, they wouldn't get those changes.  And that'd be really bad.

I feel like that's exactly what we're doing with a bunch of our Jetty settings, and it makes me nervous.

The main argument that would make me consider hard-coding our own default in for this new setting is that we're already doing it for a lot of the other settings, and as I was writing this code, it felt inconsistent because I wasn't doing it that way.  So I might be willing to change it just for the sake of consistency, but overall the pattern still feels overly risky to me.

doc/jetty-config.md
camlow325: Maybe could just add a line like...

Note that Jetty will not automatically close the connection if the idle timeout is reached while Jetty is still actively processing a client request.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
cprice404: Totally agree with you on this; the test feels sketchy to me but I thought it was probably better than nothing.  Would definitely want to remove or re-work it if we did hit any issues with it on Travis, etc.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
cprice404: I don't think I follow; if the server used a 30 second timeout on this test, the client timeout would fire at 750ms and a SocketTimeoutException would be thrown, so this test would fail... right?

doc/jetty-config.md
cprice404: Cool, will do.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
camlow325: Hmmm, yeah, I think you're right.  Since this is by default a "keep-alive" request, the server "shouldn't" close the connection right after delivering the response and so, yeah, it would seem that the client timeout would fire sooner and the test would fail.

So maybe the only potential failures might be timing-related like with the previous test.  Like if the bytes coming back from the server might be really sluggish, it could be possible for a significantly longer timeout on the server to fire while the client is still slowly receiving bytes and not yet at its limit.  Trickling a byte at a time up to 30 seconds for a localhost connection, though, seems really unlikely.

I can't think of anything to significantly improve upon the timing for this kind of test, so I'm good riding with it for now and revisiting later if we see failures in CI.

cprice404: Cool.  Will definitely rework them in some way if they are causing any issues in CI.

cprice404:
@camlow325 - pushed a commit with updated docs.  On the fence about your second bullet, maybe we wait for another review.  Third bullet point may be addressed by our comment thread?  And I agree with your fourth bullet and will add another test for that, so, at least one more commit forthcoming.

cprice404:
added a commit with the additional tests.  so AFAIK this should be done, with the possible exception of additional discussion around whether or not to hard-code a default.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
```diff
@@ -301,3 +315,103 @@
...
+        read-response (fn [client-so-timeout]
+                        (let [s (Socket. "localhost" 9000)
+                              out (PrintWriter. (.getOutputStream s) true)]
+                          (.setSoTimeout s client-so-timeout)
+                          (.println out (str "GET " path " HTTP/1.1\n"
+                                             "Host: localhost\n"
+                                             "\n"))
+                          (let [in (BufferedReader. (InputStreamReader. (.getInputStream s)))]
+                            (read-lines in))))]
+    (let [config {:webserver {:port 9000
```
fpringvaldsen: Would it be worth combining this `let` into the previous one? It seems weird to me to have two nested `let` calls one after the other.

fpringvaldsen:
I'm :+1: on this, I just had one minor comment.

In regards to setting our own default: I think that might be a good idea to keep things consistent with the other server settings we've been adding recently, since we are also setting defaults on those, and so we have a guarantee that if Jetty changes the default behavior we won't be affected.

I don't have a strong opinion on this, though, so I'm fine leaving it the way it is.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
KevinCorcoran: We've had this same debate on other PRs in the past, about other settings within Jetty.  I've weighed-in on this before, but since it's come up again ...

I'm with @cprice404 on this one.  To me, this boils down to the notion that Jetty knows better defaults for Jetty than we do.

I'd like to see us put this issue to bed for good ... either decide to continue coding our own defaults, and change this PR, or decide to stop coding our own defaults, leave this as-is, and come back around and remove all of the defaults we've already put in there.

doc/jetty-config.md
KevinCorcoran: If we're not going to hard-code our own default, I don't think we should document Jetty's default ... otherwise, we would need to figure out if any of Jetty's default configuration values had changed every time we upgrade our Jetty dependency.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
```diff
@@ -301,3 +315,103 @@
...
+                            :so-linger-milliseconds 42
+                            :idle-timeout-milliseconds 9002
+                            :client-auth :want}
+                    :jmx-enable false
+                    :max-threads 100
+                    :queue-max-size 101})
+          connectors (.getConnectors server)]
+      (is (= 2 (count connectors))
+          "Unexpected number of connectors for server")
+      (is (= 25 (.getPort (first connectors)))
```
KevinCorcoran: Are they guaranteed to come back in this order?  I _think_ `(first connectors)` here could return the other connector, in which case this test would fail incorrectly.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
```diff
@@ -301,3 +315,103 @@
...
+        path "/hi_world"
+        ring-handler (fn [req] {:status 200 :body body})
+        read-response (fn [client-so-timeout]
+                        (let [s (Socket. "localhost" 9000)
+                              out (PrintWriter. (.getOutputStream s) true)]
+                          (.setSoTimeout s client-so-timeout)
+                          (.println out (str "GET " path " HTTP/1.1\n"
+                                             "Host: localhost\n"
+                                             "\n"))
+                          (let [in (BufferedReader. (InputStreamReader. (.getInputStream s)))]
```
KevinCorcoran: I think you can replace `(BufferedReader. (InputStreamReader. (.getInputStream s)))` with `(clojure.java.io/reader (.getInputStream s))`

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
```diff
@@ -301,3 +315,103 @@
...
+              (is (thrown-with-msg? SocketTimeoutException #"Read timed out"
+                                    (read-response 250))))
+            (testing "Verify that server closes the socket after idle timeout"
+              ;; if we set the client socket timeout higher than the server,
+              ;; then the server should close the socket after its timeout,
+              ;; which will cause our read to stop blocking and allow us to
+              ;; validate the contents of the data we read from the socket.
+              (let [resp (read-response 750)]
+                (is (re-find #"(?is)HTTP.*200 OK.*Hi World"
+                             resp))))))))))
```
KevinCorcoran: Man, this is a hell of a test.  I actually think it's pretty good, re: race conditions ... at least, we've done much worse in the not-at-all-distant past.  So, that's ... something.

KevinCorcoran:
:+1: Looks mostly good to me.  Had only one non-nitpicky comment (about ordering in the test).  Aside from, of course, the issue about default settings.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
camlow325: I'm mostly in what you'd probably call the "other camp" on this one.

I totally agree that when exposing a new Jetty setting in tk-jetty9 configuration that we should draft from what Jetty is using as its internal default.  Unless we were to have a very solid reason to do otherwise, I agree that it's best to trust Jetty's judgement on what the reasonable starting point would be.

Where my opinion differs from others is around what should happen if the underlying default in Jetty were to change from one Jetty revision to another as we upgrade the tk-jetty9's Jetty dependency.  I fear that applications dependent upon the default value may be adversely affected by the default being changed on an upgrade.  For example, if core Jetty were to decide to change the `idle-timeout` from 30 seconds to 15 seconds in a later release and we were to let that new default flow through tk-jetty9 as-is, I could see the possibility for applications that have an implicit dependency on the 30 second behavior start having problems - e.g., more timed out connections - as the upgrade occurs.

Ideally such applications would isolate themselves from this risk by setting an explicit override for the settings where those applications have domain knowledge that would be more relevant than the Jetty default, e.g., an application serving requests in specific environments where it is known that the typical connection negotiation is right around the 30 second window.  I doubt it is all that common that those applications set such overrides out of the box, though.  I think it is more likely for such a setting to be updated when a problem occurs and getting to the understanding that updating the setting would fix a specific problem can be very time-consuming.

For the sake of consistency with how other settings have been layered in tk-jetty9 so far, I'd lean toward coding a default into tk-jetty9 for this PR - even though I could see this being changed in the future should the argument for trusting Jetty's defaults prevail.  Regardless of how this argument shakes out, I agree that our long-term goal should be for consistency in approach in how we layer these settings across the service implementation, even if getting to that involves work beyond the scope of this PR.

If the majority is in favor of deferring to Jetty's default for this PR, I think @KevinCorcoran raises a good point about how we want to reflect Jetty's default in external documentation.  I really like having documentation which states explicitly what the application's default is -- if nothing else than for understanding in the context of troubleshooting.  If we do document a default provided by Jetty, however, I think we need to be clear that in doing so we:

1) Are not committing to that default being preserved across upgrades.

2) Are accurately reflecting the value that Jetty is using.

Pertaining to the second point, I'd be concerned that if we don't have any tests in place to validate what Jetty is using as a default that we'd run the risk of upgrading the tk-jetty9 dependency on Jetty, not noticing that the effective default had changed, and, therefore, not updating the documentation to have the correct value.

If we decide to go with letting Jetty's default flow through for this PR, I think the simplest thing to do would be to remove the documentation about what the default is, per @KevinCorcoran's suggestion.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
camlow325: This is a good point.  The way our code path into Jetty currently works, they will come back in this order.  But there aren't any safeguards in place to ensure this will always be the case.

This test follows the pattern of a couple of a couple of other tests in this block.  It might be good to make these tests a little more robust to allow for the connectors to be enumerated back in either order.  Not sure I'd block this PR to address that issue - maybe cover in a separate PR?  I'll defer to @cprice404 on this one...

KevinCorcoran: Oh, I see - it's because we explicitly add the HTTP connector first, and then the HTTPS connector.  And `getConnectors` returns an array.  That's good enough for me.

cprice404: also this test is copypasta from jeremy's test, which has been working fine for some time.

camlow325: > also this test is copypasta from jeremy's test

That's usually not a good thing.  :smiley:

cprice404:
I don't see any outstanding comments here that weren't addressed in the threads or in the meeting earlier today, sans two nitpicks in the test code... which are valid, but I'm inclined to say they're not worth holding this up for any longer.  Objections?  Or can someone merge this?

KevinCorcoran:
:shipit:

@camlow325 I pass the ball to you.

camlow325:
:shipit:

PR #111, opened 2015-03-06
(MAINT) Prep for 1.2.0 release

cprice404:


CHANGELOG.md
```diff
@@ -1,3 +1,26 @@
+## 1.2.0
+
+This is a feature release.
+
+* Upgrade to version 9.2.8 of upstream Jetty (we were previously at
+  v9.1.0, which was over a year old).
```
KevinCorcoran: Is there any more-specific reason for the upgrade?

KevinCorcoran:
:+1: but if there is anything else that can be said about the reason for upgrading the Jetty dependency, it should be in the CHANGELOG.

cprice404:
@KevinCorcoran  Added a little more verbiage.

dankreek:
:+1:

PR #112, opened 2015-03-13
(TK-168) remove hardcoded jetty defaults

cprice404:


src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -17,7 +16,53 @@
...
+
+;;;
+;;; Host/port settings
+;;;
+;;; These are really common and fairly benign, and removing them would probably
+;;; only serve to make the bare configuration more onerous.
+;;;
+(def default-http-port 8080)
+(def default-https-port 8081)
+(def default-host "localhost")
```
dankreek: Is this set to `localhost` instead of `0.0.0.0` to ensure a webservice can't be automatically accessed by the whole Internet?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -431,19 +414,19 @@
...
         (let [client (proxy-super createHttpClient)
-              timeout (if idle-timeout (* 1000 idle-timeout)
-                                       default-proxy-idle-timeout)]
+              timeout (when idle-timeout
+                        (* 1000 idle-timeout))]
           (if (:follow-redirects options)
             (.setFollowRedirects client true)
             (.setFollowRedirects client false))
-          (.setIdleTimeout client timeout)
+          (if timeout
```
dankreek: Should maybe be `(when)` if idiomaticity is a concern.

test/clj/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_test.clj
```diff
@@ -15,9 +15,11 @@
                      with-app-with-config]]
             [puppetlabs.trapperkeeper.testutils.logging
              :refer [with-test-logging]]
-            [schema.core :as schema]))
+            [schema.core :as schema]
+            [schema.test :as schema-test]))
 
 (use-fixtures :once ks-test-fixtures/with-no-jvm-shutdown-hooks)
+(use-fixtures :once schema-test/validate-schemas)
```
nwolfe: You don't need a separate call to `use-fixtures`, you can just add `schema-test/validate-schemas` to the one above.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -16,17 +16,19 @@
...
             [puppetlabs.trapperkeeper.testutils.logging
-              :refer [with-test-logging with-log-output logs-matching]]
+             :refer [with-test-logging with-log-output logs-matching]]
             [puppetlabs.trapperkeeper.services.webserver.jetty9-core :as core]
-            [schema.core :as schema]))
+            [schema.core :as schema]
+            [schema.test :as schema-test]))
 
 (use-fixtures :once ks-test-fixtures/with-no-jvm-shutdown-hooks)
+(use-fixtures :once schema-test/validate-schemas)
```
nwolfe: Same comment...can just add to the existing `use-fixtures` instead of making another call

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -589,7 +572,8 @@
                                  (.setHandler maybe-logged))
                                maybe-logged)]
       (.setHandler s statistics-handler)
-      (.setStopTimeout s (or shutdown-timeout default-graceful-stop-timeout))
+      (if shutdown-timeout
```
dankreek: Same as above... I'm going to stop commenting on this now.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_default_config_test.clj
```diff
@@ -0,0 +1,67 @@
+;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
```
dankreek: Would it make sense to stick this big fat warning inside the namespace's docstring?

dankreek:
Over looks good to me. I only had a couple nitpicks that don't really require addressing if we're on a short time line. :+1:

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
cprice404: Yeah; and `localhost` is what we've always used here, didn't see a great reason to change it.

dankreek: Cool, that sounds reasonable. Just wanted to double check that was desired.

cprice404:
added a commit to make those things more shiny

KevinCorcoran:
Nitpicking aside, this looks :+1:   Only other comment is that there is a commit whose message starts with "WIP ..." that would probably be good to squash.

@camlow325, @nwolfe - do y'all want to review this any more?

cprice404:
whoops, yep, definitely meant to squash that WIP commit.  Will do once this has quorum.  And while I'm doing that I'll check to see if removing that key from both expected and actual causes failures or not; if it doesn't, I'll get rid of it.

README.md
```diff
@@ -314,7 +314,7 @@ route:
 * `:idle-timeout`: optional; sets the maximum amount of time, measured in seconds, for the proxy to 
   wait for a response from the upstream server. If no response is received within the time
   specified, then an HTTP 504 error is returned. If this option is not specified then a value
-  of 60 seconds is used.
+  of 0 is used, which means that the connection does not ever timeout.
```
camlow325: Where did you get '0' from?  It looked like it might be 30 seconds from here - https://github.com/eclipse/jetty.project/blob/jetty-9.2.9.v20150224/jetty-proxy/src/main/java/org/eclipse/jetty/proxy/AbstractProxyServlet.java#L268-L271.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -17,7 +16,53 @@
...
+;;; Also note that w/rt the default list of acceptable ciphers, we're deliberately
+;;; excluding all diffie-helman ciphers due to some old JDK bugs:
+;;;
+;;; http://bugs.java.com/bugdatabase/view_bug.do?bug_id=8014618
+;;; https://github.com/puppetlabs/puppetdb/commit/03e020dc85b83d6c83c9992ca6bd14f57e8fc91a
+;;;
+;;; These have been fixed in recent versions of the JDK, and it would be nice
+;;; to be able to add the DH ciphers back in at some point, but we can't do that
+;;; until we're certain that our minimum supported JDK versions for all of our
+;;; supported distros will contain the relevant fixes.
```
camlow325: Yep, good description.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_default_config_test.clj
```diff
@@ -0,0 +1,67 @@
...
+  (:import (org.eclipse.jetty.server HttpConfiguration ServerConnector Server)
+           (org.eclipse.jetty.client HttpClient)))
+
+(use-fixtures :once schema-test/validate-schemas)
+
+(deftest default-request-header-size-test
+  (let [http-config (HttpConfiguration.)]
+    (is (= 8192 (.getRequestHeaderSize http-config)))))
+
+(deftest default-jetty-http-client-settings-test
```
camlow325: Presumably, these tests are here to validate the defaults that the ProxyServlet uses for the HttpClient it creates.  I don't think that will work, though, because the ProxyServlet applies its own defaults over the top of the ones that the HttpClient constructor creates.  See: https://github.com/eclipse/jetty.project/blob/jetty-9.2.9.v20150224/jetty-proxy/src/main/java/org/eclipse/jetty/proxy/AbstractProxyServlet.java#L268-L271 and https://github.com/eclipse/jetty.project/blob/jetty-9.2.9.v20150224/jetty-proxy/src/main/java/org/eclipse/jetty/proxy/AbstractProxyServlet.java#L278-L280.  This may work okay for the `requestBufferSize` for now since the ProxyServlet doesn't change the default but it doesn't appear that this will work for `idleTimeout` because the default is changed to 30 seconds.

We may need to have these tests construct a ProxyServlet and read the defaults back from what the `createHttpClient()` method would return - or just read back the _client directly (although that's a private member variable of the AbstractProxyServlet class so we'd probably have to do some ugly reflection to do that).

camlow325:
:+1: other than the questions about the ProxyServlet-related defaults.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
```diff
@@ -27,192 +30,218 @@
...
-    (is (expected-http-config?
-          {:port 8000 :queue-max-size 123}
-          {:http {:host default-host :port 8000}
-           :queue-max-size 123}))
-
-    (is (expected-http-config?
-          {:port 8000 :idle-timeout-milliseconds 6000}
-          {:http {:host default-host :port 8000
-                  :idle-timeout-milliseconds 6000}}))))
+    (is (= (munge-actual-http-config
```
camlow325: This was a great addition.  Thanks much!

README.md
cprice404: Got it from my test, which as you pointed out might not be testing the right thing.  Will try to sort that out.

cprice404:
Rebased onto @camlow325 's latest changes (bumps to Jetty 9.2.10), squashed the WIP commit that @KevinCorcoran pointed out, and fixed the defaults tests around the ProxyServlet client settings.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_default_config_test.clj
```diff
@@ -0,0 +1,89 @@
...
+            [puppetlabs.trapperkeeper.services.webserver.jetty9-core :as core])
+  (:import (org.eclipse.jetty.server HttpConfiguration ServerConnector Server)))
+
+(use-fixtures :once schema-test/validate-schemas)
+
+(deftest default-request-header-size-test
+  (let [http-config (HttpConfiguration.)]
+    (is (= 8192 (.getRequestHeaderSize http-config)))))
+
+(deftest default-proxy-http-client-settings-test
```
camlow325: Nice - works for me.

camlow325:
:+1: assuming Travis passes.  Sounds like @KevinCorcoran is good with it as well, so I'll merge after that if no problems come up.

PR #113, opened 2015-03-17
(TK-178) Upgrade Jetty dependency to 9.2.10

camlow325:
This commit upgrades the tk-jetty9's Jetty dependency to 9.2.10.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,5 +1,6 @@
 (def tk-version "1.1.0")
 (def ks-version "1.0.0")
+(def jetty-version "9.2.10.v20150310")
 
 (defproject puppetlabs/trapperkeeper-webserver-jetty9 "1.2.1-SNAPSHOT"
   :description "A jetty9-based webserver implementation for use with the puppetlabs/trapperkeeper service framework."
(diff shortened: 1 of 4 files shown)


KevinCorcoran:
Would be good to document why ... maybe the CHANGELOG is the correct place for that.

camlow325:
Yeah, I figured we'd document that in the CHANGELOG.md when we do the next release.

project.clj
```diff
@@ -25,13 +25,13 @@
...
-                 [org.eclipse.jetty/jetty-servlet "9.2.8.v20150217"]
-                 [org.eclipse.jetty/jetty-servlets "9.2.8.v20150217"]
-                 [org.eclipse.jetty/jetty-webapp "9.2.8.v20150217"]
-                 [org.eclipse.jetty/jetty-proxy "9.2.8.v20150217"]
-                 [org.eclipse.jetty/jetty-jmx "9.2.8.v20150217"]
+                 [org.eclipse.jetty/jetty-servlet "9.2.10.v20150310"]
+                 [org.eclipse.jetty/jetty-servlets "9.2.10.v20150310"]
+                 [org.eclipse.jetty/jetty-webapp "9.2.10.v20150310"]
+                 [org.eclipse.jetty/jetty-proxy "9.2.10.v20150310"]
+                 [org.eclipse.jetty/jetty-jmx "9.2.10.v20150310"]
```
nwolfe: Would it be worthwhile to `def` this version string? They're all the same, and all have to change at the same time right?

camlow325: That thought has crossed my mind a few times.  Not sure if we did it this way just so we were more intentional about understanding what deps we were going to be pulling in.  Each time I've updated Jetty, I've downloaded the Jetty source and run mvn deps on it to confirm that all of these match up for a given release.  Maybe I've been too paranoid there.

I'm open to this change, though.  @cprice404 - what do you think?

nwolfe:
:+1: other than minor version string comment

project.clj
cprice404: i'm fine either way

camlow325:
I went ahead and did the common def in af64648.

PR #114, opened 2015-03-20
(TK-148) Expose selectors and acceptors and adjust thread defaults

camlow325:
This commit contains three primary changes:

1) Expose the ability to configure the number of `selector-threads` and
   `acceptor-threads` that the webserver will spin up per "port" in a
   server configuration.

2) Remove the library-imposed defaults for various thread-pool related
   settings, including: acceptor-threads, selector-threads, max-threads,
   and queue-max-size.  The service will now defer to Jetty to impose
   its own default values for these settings.

3) In the process of removing library-imposed defaults, this commit
   effectively removes the changes from TK-130.  The `max-threads` setting
   will no longer automatically be bumped up above the minimum number of
   threads that the server needs to boot if the number that the server
   tries to use is insufficient and the user had not configured an explicit
   value.  Some documentation was added to the `max-threads` setting to
   advise users on how to configure the value manually, if needed to avoid
   this error.

### DIFF
--- a/doc/jetty-config.md
+++ b/doc/jetty-config.md
@@ -1,7 +1,7 @@
 ## Configuring The Webserver Service
 
-The `webserver` section in your Trapperkeeper configuration files configures an embedded
-HTTP server inside trapperkeeper.
+The `webserver` section in your Trapperkeeper configuration files configures an
+embedded HTTP server inside trapperkeeper.
 
 ### `host`
 
@@ -16,21 +16,60 @@ This sets what port to use for _unencrypted_ HTTP traffic.  If not supplied, but
 `host` is supplied, a value of 8080 will be used.  If neither host nor port is
 supplied, we won't listen for unencrypted traffic at all.
 
+### `acceptor-threads`
+
+This sets the number of threads that the webserver will dedicate to accepting
+socket connections for _unencrypted_ HTTP traffic.  Defaults to the number of
+virtual cores on the host divided by 8, with a minimum of 1 and maximum of 4.
+
+### `selector-threads`
+
+This sets the number of threads that the webserver will dedicate to processing
+events on connected sockets for _unencrypted_ HTTP traffic.  Defaults to the
+number of virtual cores on the host divided by 2, with a minimum of 1 and
+maximum of 4.
+
 ### `max-threads`
 
-This sets the maximum number of threads assigned to responding to HTTP and HTTPS
-requests, effectively changing how many concurrent requests can be made at one
-time. Defaults to 100.
+This sets the maximum number of threads assigned to responding to HTTP and/or
+HTTPS requests for a single webserver, effectively changing how many concurrent
+requests can be made at one time.  Defaults to 200.
+
+Each webserver instance requires a minimum number of threads in order to boot
+properly.  The minimum number is calculated as:
+
+~~~~
+(number of "acceptor-threads" for each port) +
+(number of "selector-threads" for each port) +
+1 "worker" thread
+~~~~
+
+"1" is the minimum number of worker threads required to process incoming
+web requests.
+
+For example, if an _unencrypted_ port with 2 `acceptor-threads` and 3
+`selector-threads` and an _encrypted_ port with 4 `acceptor-threads` and 5
+`selector-threads` were configured with the webserver, the webserver would
+require that a minimum value of 15 (2 + 3 + 4 + 5 + 1) be used for the
+`max-threads` setting.  If the configured value for `max-threads` is less
+than the minimum required value, server startup will fail with an
+`IllegalStateException`, with a message containing the words
+"insufficient threads".
+
+Note that each web request must be processed on a "worker" thread which is
+separate from the acceptor and selector threads.  The configured `max-threads`
+value should allow for the maximum number of requests which are desired for the
+server be able to handle concurrently.
 
 ### `queue-max-size`
 
 This can be used to set an upper-bound on the size of the worker queue that the
 web server uses to temporarily store incoming client connections before they
-can be serviced.  This value defaults to "unbounded".  A request which is
-rejected by the web server because the queue is full would be seen by the
-client as having initially connected to the server socket at the TCP layer but
-having been closed almost immediately afterward by the server with no HTTP
-layer response body.
+can be serviced.  This value defaults to the maximum value of a 32-bit signed
+integer, 2147483647.  A request which is rejected by the web server because the
+queue is full would be seen by the client as having initially connected to the
+server socket at the TCP layer but having been closed almost immediately
+afterward by the server with no HTTP layer response body.
 
 ### `request-header-max-size`
 
@@ -70,6 +109,19 @@ This sets the port to use for _encrypted_ HTTPS traffic. If not supplied, but
 `ssl-host` is supplied, a value of 8081 will be used for the https port.  If
 neither ssl-host nor ssl-port is supplied, we won't listen for encrypted traffic
 at all.
+
+### `ssl-acceptor-threads`
+
+This sets the number of threads that the webserver will dedicate to accepting
+socket connections for _encrypted_ HTTPS traffic.  Defaults to the number of
+virtual cores on the host divided by 8, with a minimum of 1 and maximum of 4.
+
+### `ssl-selector-threads`
+
+This sets the number of threads that the webserver will dedicate to processing
+events on connected sockets for _encrypted_ HTTPS traffic.  Defaults to the
+number of virtual cores on the host divided by 2, with a minimum of 1 and
+maximum of 4.
 
 ### `ssl-cert`
 
--- a/doc/jetty-config.md
+++ b/doc/jetty-config.md
@@ -44,9 +44,6 @@ properly.  The minimum number is calculated as:
 1 "worker" thread
 ~~~~
 
-"1" is the minimum number of worker threads required to process incoming
-web requests.
-
 For example, if an _unencrypted_ port with 2 `acceptor-threads` and 3
 `selector-threads` and an _encrypted_ port with 4 `acceptor-threads` and 5
 `selector-threads` were configured with the webserver, the webserver would
@@ -57,9 +54,11 @@ than the minimum required value, server startup will fail with an
 "insufficient threads".
 
 Note that each web request must be processed on a "worker" thread which is
-separate from the acceptor and selector threads.  The configured `max-threads`
-value should allow for the maximum number of requests which are desired for the
-server be able to handle concurrently.
+separate from the acceptor and selector threads.  "1" is the minimum number of
+worker threads required to process incoming web requests.  The `max-threads`
+value should be large enough that the server can allocate all of the selector
+and acceptor threads that it needs and yet still have a sufficient number of
+worker threads left over for handling concurrent web requests.
 
 ### `queue-max-size`
 
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
@@ -81,22 +81,6 @@
 ;;; leaving them on by default.
 (def default-jmx-enable "true")
 
-
-;; TODO: these two need to be addressed in our upcoming work around
-;; acceptors/selectors.  See TK-148.
-(def default-max-threads 100)
-(def default-queue-max-size (Integer/MAX_VALUE))
-;; TODO: these need to be taken care of too; I moved them here from core
-;; just to try to get all this stuff into one spot for cleaning up in TK-148.
-(def default-queue-idle-timeout
-  "The maximum number of milliseconds that a thread in the queue can remain
-  idle before the thread may be thrown away.  A value less than or equal to 0
-  would allow a thread to remain idle indefinitely."
-  60000)
-(def default-queue-min-threads
-  "The minimum number of threads to create and maintain in the queue."
-  8)
-
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Schemas
 
@@ -108,6 +92,8 @@
 (def WebserverRawConfig
   {(schema/optional-key :port)                       schema/Int
    (schema/optional-key :host)                       schema/Str
+   (schema/optional-key :acceptor-threads)           schema/Int
+   (schema/optional-key :selector-threads)           schema/Int
    (schema/optional-key :max-threads)                schema/Int
    (schema/optional-key :queue-max-size)             schema/Int
    (schema/optional-key :request-header-max-size)    schema/Int
@@ -119,6 +105,8 @@
    (schema/optional-key :ssl-cert)                   schema/Str
    (schema/optional-key :ssl-cert-chain)             schema/Str
    (schema/optional-key :ssl-ca-cert)                schema/Str
+   (schema/optional-key :ssl-acceptor-threads)       schema/Int
+   (schema/optional-key :ssl-selector-threads)       schema/Int
    (schema/optional-key :keystore)                   schema/Str
    (schema/optional-key :truststore)                 schema/Str
    (schema/optional-key :key-password)               schema/Str
@@ -181,7 +169,9 @@
 (def WebserverConnector
   (merge WebserverConnectorCommon
          {:host schema/Str
-          :port schema/Int}))
+          :port schema/Int
+          :acceptor-threads (schema/maybe schema/Int)
+          :selector-threads (schema/maybe schema/Int)}))
 
 (def WebserverSslContextFactory
   {:keystore-config                    WebserverSslKeystoreConfig
@@ -205,8 +195,8 @@
     HasConnector
     {(schema/optional-key :http)  WebserverConnector
      (schema/optional-key :https) WebserverSslConnector
-     :max-threads                 schema/Int
-     :queue-max-size              schema/Int
+     :max-threads                 (schema/maybe schema/Int)
+     :queue-max-size              (schema/maybe schema/Int)
      :jmx-enable                  schema/Bool}))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
@@ -387,8 +377,10 @@
   [config :- WebserverRawConfig]
   (if (contains-http-connector? config)
     (merge (common-connector-config config)
-           {:host (or (:host config) default-host)
-            :port (or (:port config) default-http-port)})))
+           {:host             (or (:host config) default-host)
+            :port             (or (:port config) default-http-port)
+            :acceptor-threads (:acceptor-threads config)
+            :selector-threads (:selector-threads config)})))
 
 (schema/defn ^:always-validate
   contains-https-connector? :- schema/Bool
@@ -402,6 +394,8 @@
     (merge (common-connector-config config)
            {:host                    (or (:ssl-host config) default-host)
             :port                    (or (:ssl-port config) default-https-port)
+            :acceptor-threads        (:ssl-acceptor-threads config)
+            :selector-threads        (:ssl-selector-threads config)
             :keystore-config         (get-keystore-config! config)
             :cipher-suites           (get-cipher-suites-config config)
             :protocols               (get-ssl-protocols-config config)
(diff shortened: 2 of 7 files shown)


src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -81,22 +81,6 @@
 ;;; leaving them on by default.
 (def default-jmx-enable "true")
 
-
```
camlow325: These were all removed since we're deferring to Jetty's defaults for these now.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -433,59 +427,6 @@
     (throw (IllegalArgumentException.
              "Error: More than one default server specified in configuration"))))
 
-(defn selectors-count
```
camlow325: Production code wouldn't use these anymore however the count functions are still being used in the tests in the jetty9-default-config-test namespace to help validate that Jetty is using the algorithm that we think it is for calculating acceptor and selector threads.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -270,25 +273,34 @@
...
-                         queue-max-size))))
+  queue-thread-pool :- (schema/maybe QueuedThreadPool)
+  [max-threads :- (schema/maybe schema/Int)
+   queue-max-size :- (schema/maybe schema/Int)]
+  (if (or max-threads queue-max-size)
+    (let [thread-pool (if max-threads
+                        (QueuedThreadPool. max-threads)
+                        (QueuedThreadPool.))]
+      (if queue-max-size
+        (let [min-threads    (.getMinThreads thread-pool)
```
camlow325: I hated having to do the double-construction of a `QueuedThreadPool` but couldn't think of a better way to configure Jetty to use a custom value for `queue-max-size` while still preserving the other defaults that Jetty uses for other settings in the thread pool configuration.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -495,9 +436,11 @@
   (let [result (-> {}
                    (maybe-add-http-connector config)
                    (maybe-add-https-connector config)
-                   (assoc :max-threads (determine-max-threads config (num-cpus)))
```
camlow325: We would no longer be auto-adjusting `max-threads` up to the needed server minimum - prior work from TK-130.  The assumption is that the user would now have to set `max-threads` manually in the case that the number of acceptor and selector threads that the server needs to use would exceed the default `max-threads`.  This probably wouldn't be common anyway but is now documented in the `max-threads` section of the Jetty config doc.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
```diff
@@ -372,129 +409,6 @@
       "./dev-resources/config/jetty/ssl/certs/master-with-intermediate-ca.pem"
       "./dev-resources/config/jetty/ssl/certs/ca-root.pem")))
 
-(deftest determine-max-threads-test
```
camlow325: These are all gone now because they were only needed to validate the auto-adjustment of `max-threads` from TK-130, which would no longer exist per this PR.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
```diff
@@ -189,125 +192,139 @@
...
-  {:http {:port 0
-          :host "0.0.0.0"
-          :request-header-max-size 100
-          :so-linger-milliseconds so-linger-milliseconds
-          :idle-timeout-milliseconds idle-timeout-milliseconds}
-   :jmx-enable false
-   :max-threads max-threads
-   :queue-max-size queue-max-size})
-
-(defn get-thread-pool-and-queue-for-create-server
```
camlow325: The guts of this moved over to `get-server-thread-pool-queue` in the jetty9-default-config-test namespace, since it now has a similar need for this.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
```diff
@@ -189,125 +192,139 @@
...
-        ;; the thread pool.
-        get-queue-method (-> thread-pool
-                             (.getClass)
-                             (.getDeclaredMethod "getQueue" nil))
-        _                (.setAccessible get-queue-method true)
-        queue            (.invoke get-queue-method thread-pool nil)]
-    {:thread-pool thread-pool
-     :queue queue}))
-
-(deftest create-server-thread-pool-test
```
camlow325: These tests still effectively exist under `create-server-queue-max-size-test` but they've been refactored to eliminate some duplication and also validate Jetty defaults when not explicitly set in user configuration.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
cprice404: Hmm.. so this is just for 'getMaxThreads' and 'getIdleTimeout'?  Do either of those involve any fancy computations or are they just hard-coded numeric values in Jetty?

If the latter, I'm thinking maybe it would be a lesser evil to just put those hard coded values up into our defaults section with some comments explaining why they're there, and a reference to the tests that validate them... and then a comment next to those tests explaining that if they fail in the future then the user needs to update the defaults?  Or maybe the tests should just be written to use those hard-coded vars in their implementation?

I know that goes against the grain of what we've been trying to do here, but it seems maybe slightly less bad than having to construct a thread pool just to throw it away?

camlow325: Most of it is just hard-coded numbers in Jetty - see https://github.com/eclipse/jetty.project/blob/jetty-9.2.10.v20150310/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java#L65-L98.  The only part that's somewhat more interesting is the max (_minThreads, 8) part, which the line below this in the Clojure code mimics.  The assumption being that the "8" hard-coded in two places in the Jetty code is representative of an intention to tie min-threads to the initial capacity of the thread pool's queue.

I'm not too worried about the expense of the double-construction.  Appears to mostly be setting up primitive values for fields in the `QueuedThreadPool` object.  The backing `BlockingArrayQueue` object will allocate an Object array per the initial capacity of the queue, which won't be larger than 8 - see https://github.com/eclipse/jetty.project/blob/jetty-9.2.10.v20150310/jetty-util/src/main/java/org/eclipse/jetty/util/BlockingArrayQueue.java#L113-L118.  The expensive operations wouldn't happen until after the thread pool is started, which it never would be for the first one allocated here.

I think I'd still prefer the double-construction to hard coding defaults in our code since this is a one-time thing at startup and is pretty inexpensive.

cprice404: If you're sure it doesn't actually start creating threads then I don't have too big of a problem with this implementation, depending on what other folks think.

If we do keep it this way I think you should add a comment though, so that people don't think you're nuts when they look at this in the future.

cprice404:
Overall I'm a big +1 on this.  Lots of deletions.  yay :)

I only reviewed the tests at a high level, it's a bit tricky to review since stuff is moving around, but it generally looks good to me.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
camlow325: Yeah, none of the threads would start unless `start` were called on the `QueuedThreadPool` - https://github.com/eclipse/jetty.project/blob/jetty-9.2.10.v20150310/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java#L106 - which would never happen on the first one constructed, in this case.

I'll add a comment to the code, but I guarantee people will think I'm nuts no matter what.  :smile:

camlow325:
I added a comment about the extra `QueuedThreadPool` construction for the `queue-max-size` case.  That's in commit 2249af4.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -270,25 +273,50 @@
...
+        ;; `queue-max-size` were configured.
+        (let [min-threads    (.getMinThreads thread-pool)
+              queue-min-size (min queue-max-size min-threads)]
+          (QueuedThreadPool. (.getMaxThreads thread-pool)
+                             min-threads
+                             (.getIdleTimeout thread-pool)
+                             (BlockingArrayQueue.
+                               queue-min-size
+                               queue-min-size
+                               queue-max-size)))
```
KevinCorcoran: I'm okay with the approach you took here (creating the extra thread pool) and your comment is helpful.  I think these changes are fine, however, I want to raise a question unrelated to your changes ... 

Why are we setting the min size and growth raise on the queue?  It seems like all we really care about is the max size, and there is a constructor that takes only the max size which we could use instead.  Furthermore, the calculation for min size doesn't quite make sense, and it's also quite confusing to me that we're setting growth rate == min size.

doc/jetty-config.md
```diff
@@ -16,21 +16,60 @@ This sets what port to use for _unencrypted_ HTTP traffic.  If not supplied, but
...
+require that a minimum value of 15 (2 + 3 + 4 + 5 + 1) be used for the
+`max-threads` setting.  If the configured value for `max-threads` is less
+than the minimum required value, server startup will fail with an
+`IllegalStateException`, with a message containing the words
+"insufficient threads".
+
+Note that each web request must be processed on a "worker" thread which is
+separate from the acceptor and selector threads.  The configured `max-threads`
+value should allow for the maximum number of requests which are desired for the
+server be able to handle concurrently.
```
KevinCorcoran: I think this last sentence needs re-wording, it's hard to understand as written.

Also, crazy question here, but, since we're making acceptor + selector thread counts configurable, should we do the same for the worker threads?

doc/jetty-config.md
```diff
@@ -16,21 +16,60 @@ This sets what port to use for _unencrypted_ HTTP traffic.  If not supplied, but
...
+properly.  The minimum number is calculated as:
+
+~~~~
+(number of "acceptor-threads" for each port) +
+(number of "selector-threads" for each port) +
+1 "worker" thread
+~~~~
+
+"1" is the minimum number of worker threads required to process incoming
+web requests.
```
KevinCorcoran: I think you should move this sentence into the last paragraph where you are also talking about worker threads.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_default_config_test.clj
```diff
@@ -71,19 +75,110 @@ react accordingly."
...
+          (insufficient-threads-msg [server]
+            (let [connectors (count (.getConnectors server))]
+              (re-pattern (str "Insufficient threads: max="
+                               (get-max-threads-for-server server)
+                               " < needed\\(acceptors="
+                               (* acceptor-thread-count connectors)
+                               " \\+ selectors="
+                               (* selector-thread-count connectors)
+                               " \\+ request=1\\)"))))]
+    (dotimes [x 5]
```
KevinCorcoran: why 5 times?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
camlow325: If we go with the `BlockingArrayQueue` constructor that just takes a max capacity, we'd end up with a fixed-sized queue that's pre-allocated up to the maximum capacity specified.  See https://github.com/eclipse/jetty.project/blob/jetty-9.2.10.v20150310/jetty-util/src/main/java/org/eclipse/jetty/util/BlockingArrayQueue.java#L101.

The Server's default `QueuedThreadPool`, however, creates a `BlockingArrayQueue` that is allocated with a minimum capacity with the ability to grow up toward the maximum.  See: https://github.com/eclipse/jetty.project/blob/jetty-9.2.10.v20150310/jetty-util/src/main/java/org/eclipse/jetty/util/thread/QueuedThreadPool.java#L92-L96.  I was thinking we'd want to preserve the "grow" capability to keep memory usage in check and behave more like the way the server would by default.

`QueuedThreadPool` sets the min size and growth rate based on the maximum of min threads in the thread pool or "8".  The "8" here also corresponds to the default value for min threads in the pool.  Since we don't allow for min threads in the pool to be configured, I just went with fixing this to min threads in the pool only.

The `(min queue-max-size min-threads)` is something Jetty doesn't protect for but that seemed worthwhile.  `BlockingArrayQueue` will throw an IllegalArgumentException if the initial capacity is larger than the maximum capacity - see https://github.com/eclipse/jetty.project/blob/jetty-9.2.10.v20150310/jetty-util/src/main/java/org/eclipse/jetty/util/BlockingArrayQueue.java#L133.  Since we don't expose a way to configure initial capacity, it seemed reasonable to me to just head off the possibility of the IllegalArgumentException by adjusting the initial capacity to be equal to maximum capacity.  I suspect this to be pretty unlikely anyway to set such a small max capacity for the queue for anything but testing.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_default_config_test.clj
```diff
@@ -71,19 +75,110 @@ react accordingly."
...
+(defn get-server-thread-pool-queue
+  [server]
+  (let [thread-pool      (.getThreadPool server)
+        ;; Using reflection here because the .getQueue method is protected and I
+        ;; didn't see any other way to pull the queue back from the thread pool.
+        get-queue-method (-> thread-pool
+                             (.getClass)
+                             (.getDeclaredMethod "getQueue" nil))
+        _                (.setAccessible get-queue-method true)]
+    (.invoke get-queue-method thread-pool nil)))
```
KevinCorcoran: this is pretty evil!  :clap:

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
```diff
@@ -189,125 +192,139 @@
...
+      (is (= (.getMaxCapacity (get-server-thread-pool-queue (Server.)))
+             (.getMaxCapacity (get-queue-for-partial-http-config
+                                {:queue-max-size nil})))))
+    (testing "custom default queue max size passed through to thread pool queue"
+      (is (= 393
+             (.getMaxCapacity (get-queue-for-partial-http-config
+                                {:queue-max-size 393})))))
+    (testing (str "default max threads passed through to thread pool when "
+                  "queue-max-size set")
+      (is (= (default-server-max-threads)
```
KevinCorcoran: This looks like a bug - looks like accidental parens around `default-server-min-threads` - that seems like it should throw an error since `default-server-min-threads` should be an integer, right?

KevinCorcoran: oh, nevermind

doc/jetty-config.md
camlow325: Worker threads effectively are configurable.  Basically, the difference between the sums of `acceptor-threads` and `selector-threads` across connectors and `max-threads` is the maximum number of worker threads that the server will allocate.  Worker threads are different from `acceptor-threads` and `selector-threads` in they aren't fixed at boot time.  The server will only allocate one by default but can allocate additional threads, as needed to handle incoming traffic, up to the `max-threads` for the pool.  The server can also let the number of worker threads scale back down after a peak period for requests has passed.

If we hadn't already had the concept of a `max-threads` setting, it might have made more sense to me to have a `worker-threads` setting and we could use that to calculate the `max-threads` that we give to Jetty during configuration.  It might be too confusing to change that now, and it also would seem to get us into the mode of having our production code be more knowledgeable about exactly how this algorithm is used by Jetty - which I think we're trying to avoid as much as possible now.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
```diff
@@ -189,125 +192,139 @@
...
+  [config]
+  (.getThreadPool (create-server-with-partial-http-config config)))
+
+(defn get-thread-pool-for-default-server
+  []
+  (.getThreadPool (Server.)))
+
+(defn default-server-max-threads
+  []
+  (.getMaxThreads (get-thread-pool-for-default-server)))
```
KevinCorcoran: This stuff would probably be a bit more readable like...
```clj
(def default-server (Server.)))

(def default-server-max-threads (.getMaxThreads default-server)))
```

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_default_config_test.clj
camlow325: It's arbitrary.  Could just do 2 times to cover the maximum number we'd ever use for a connector, for now.  Our testing here is indirect anyway since we're throwing some numbers and Jetty and using our expectation of what it will calculate to give us some confidence that we know how their algorithm works.  I'd thought that by throwing a few more connectors in there that it might increase that confidence a bit more.  I could be convinced to bump it back down to 2, though, if you think that's better.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
KevinCorcoran: Okay, I'm on-board with setting min size.  

However, in the second link you provided above, notice that the growth rate is not specified - so I'm still not sure why we are setting that, and why we're using `queue-min-size` as the value.

Also, this seems like really good information to have in (yet another) comment in this function.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_default_config_test.clj
```diff
@@ -71,19 +75,110 @@ react accordingly."
...
+(defn calculate-minimum-required-threads
+  "Calculate the minimum number threads that a single Jetty Server instance
+  needs.  See: https://github.com/eclipse/jetty.project/blob/jetty-9.2.10.v20150310/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java#L334-L350"
+  [connectors]
+  (+ 1 (* connectors threads-per-connector)))
 
+(deftest default-min-threads-settings-test
+  ;; This test just exists to validate the advice we give for the bare
+  ;; minimum number of threads that one should account for when setting the
+  ;; 'max-threads' setting for a server instance.
```
KevinCorcoran: Which advice is this referring to?  I just re-read the docs for `max-threads` but I couldn't figure out what you meant by that.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
camlow325: > However, in the second link you provided above, notice that the growth rate is not specified - so I'm still not sure why we are setting that, and why we're using queue-min-size as the value.

Not sure I follow.  Jetty's `QueuedThreadPool`, by default, is calling the form of the `BlockingArrayQueue` constructor that takes in a "capacity" and a "growBy".  See https://github.com/eclipse/jetty.project/blob/jetty-9.2.10.v20150310/jetty-util/src/main/java/org/eclipse/jetty/util/BlockingArrayQueue.java#L113.  It sets them both to the same value, which derives from the minimum number of threads in the `QueuedThreadPool`.

I could define a `queue-capacity` and `queue-grow-by` and set them both to the result of `(min queue-max-size min-threads)` if that would help for clarity, although that wouldn't produce a functional difference.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_default_config_test.clj
KevinCorcoran: Oh, I guess you're just referring to the calculation in Jetty.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_default_config_test.clj
KevinCorcoran: I just realized that the purpose of this test is that it will fail if Jetty changes the way the calculate the minimum required # of threads ... right?

In that case ... I'm not sure why repeating this more than once makes sense.  What could possibly cause it to pass a couple of times and then eventually fail?  Like, if Jetty changed something that made the calculation different on subsequent connectors?

I gotta say, 2 seems a lot less arbitrary than 5, so I'd feel better about that.   Also I think this test could use some extra comments or `testing` strings or whatever that capture some of this info.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
KevinCorcoran: Oh, my bad, I was mistaken on what those 2 constructor params were.

I think that change would be an improvement, but I'm fine without it too - however, a comment would still be really helpful.

And thanks for talking me through this.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_default_config_test.clj
camlow325: Yeah, just the fact that Jetty calculates its min-threads based on the number of selector and acceptor threads that it needs to allocate.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_default_config_test.clj
camlow325: I'll go with 2 and add a comment that we're doing that because 2 is the maximum number of connectors our library allows to be associated with a server - at most one for plaintext and one for SSL.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
camlow325: Ok, I can do that.

camlow325:
I added commit 9211f47 to address some PR feedback:

1) Hopefully clarified the language in the `max-threads` setting definition.

2) Added some more comments around the construction of the `QueuedThreadPool` when `queue-max-size` is set.

3) Changed a couple of defns to defs for stuff that could be memoized.

4) Changed the `default-min-threads-settings-test` to only exercise one and two connector per server scenarios since those are the only ones which are possible for a tk-jetty9 server.

KevinCorcoran:
Tests passed for me locally.

PR #116, opened 2015-04-09
(TK-195) Update prismatic library dependencies

dparis:
The prismatic/plumbing and prismatic/schema library dependencies are
lagging behind the latest stable version. This will increasingly cause
conflicts in trapperkeeper projects. This patch updates the dependencies
and fixes several functions which were validating against schemas
incompatible with the new versions.

* Update project.clj dependencies
* Use protocol schema instead of the protocol object directly

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -14,7 +14,8 @@
   :dependencies [[org.clojure/clojure "1.6.0"]
                  [org.clojure/tools.logging "0.2.6"]
                  [clj-time "0.5.1"]
-                 [prismatic/schema "0.2.2"]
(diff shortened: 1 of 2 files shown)


dparis:
This is blocked by https://github.com/puppetlabs/clj-http-client/pull/39

cprice404:
The travis failure here looks like it would probably be fixed by merging the `clj-http-client` PR, doing a release of `clj-http-client`, and bumping up the dependency to point at that.

cprice404:
Going to merge this and then add a few touch-up commits to fix the travis failures.

PR #120, opened 2015-08-07
Add websockets support

richardc:
Here we add add-websocket-handler to the WebserverService and WebroutingService, along with supporting code.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
@@ -123,6 +123,14 @@
         add-servlet-handler (:add-servlet-handler webserver-service)]
     (add-servlet-handler servlet path opts)))
 
+(schema/defn ^:always-validate add-websocket-handler!
+  [context webserver-service
+   svc :- (schema/protocol tk-services/Service)
+   handlers options :- RouteOption]
+  (let [{:keys [path opts]}   (compute-common-elements context svc options)
+        add-websocket-handler (:add-websocket-handler webserver-service)]
+    (add-websocket-handler handlers path opts)))
+
 (schema/defn ^:always-validate add-war-handler!
   [context webserver-service
    svc :- (schema/protocol tk-services/Service)
--- a/src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
@@ -21,6 +21,9 @@
 (def RouteOption
   {(schema/optional-key :route-id) schema/Keyword})
 
+(def CommonOptions
+  (dissoc (merge jetty9-core/CommonOptions RouteOption) :server-id))
+
 (def ContextHandlerOptions
   (dissoc (merge jetty9-core/ContextHandlerOptions RouteOption) :server-id))
 
@@ -110,7 +113,7 @@
 (schema/defn ^:always-validate add-ring-handler!
   [context webserver-service
    svc :- (schema/protocol tk-services/Service)
-   handler options :- RouteOption]
+   handler options :- CommonOptions]
   (let [{:keys [path opts]} (compute-common-elements context svc options)
         add-ring-handler    (:add-ring-handler webserver-service)]
     (add-ring-handler handler path opts)))
@@ -126,7 +129,7 @@
 (schema/defn ^:always-validate add-websocket-handler!
   [context webserver-service
    svc :- (schema/protocol tk-services/Service)
-   handlers options :- RouteOption]
+   handlers options :- CommonOptions]
   (let [{:keys [path opts]}   (compute-common-elements context svc options)
         add-websocket-handler (:add-websocket-handler webserver-service)]
     (add-websocket-handler handlers path opts)))
--- a/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
+++ b/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
@@ -4,13 +4,16 @@
            (java.nio.file Paths Files)
            (java.nio.file.attribute FileAttribute))
   (:require [clojure.test :refer :all]
+            [gniazdo.core :as ws-client]
+            [puppetlabs.experimental.websockets.client :as ws-session]
             [puppetlabs.trapperkeeper.services.webserver.jetty9-service :refer :all]
             [puppetlabs.trapperkeeper.testutils.webserver.common :refer :all]
             [puppetlabs.trapperkeeper.app :refer [get-service]]
             [puppetlabs.trapperkeeper.testutils.bootstrap :refer [with-app-with-config]]
             [puppetlabs.trapperkeeper.testutils.logging
              :refer [with-test-logging]]
-            [schema.test :as schema-test]))
+            [schema.test :as schema-test]
+            [clojure.tools.logging :as log]))
 
 (use-fixtures :once schema-test/validate-schemas)
 
@@ -171,6 +174,57 @@
           (is (= (:status response) 200))
           (is (= (:body response) init-param-two)))))))
 
+(deftest websocket-test
+  (testing "Websocket handlers"
+    (with-app-with-config app
+      [jetty9-service]
+      jetty-plaintext-config
+      (let [s                     (get-service app :WebserverService)
+            add-websocket-handler (partial add-websocket-handler s)
+            path                  "/test"
+            connected              (atom 0)
+            server-messages        (atom [])
+            server-binary-messages (atom [])
+            client-messages        (atom [])
+            client-binary-messages (atom [])
+            binary-client-message  (promise)
+            closed                 (promise)
+            handlers               {:on-connect (fn [ws]
+                                                  (ws-session/send! ws "Hello client!")
+                                                  (swap! connected inc))
+                                    :on-text    (fn [ws text]
+                                                  (ws-session/send! ws (str "You said: " text))
+                                                  (swap! server-messages conj text))
+                                    :on-bytes   (fn [ws bytes offset len]
+                                                  (let [as-vec (vec bytes)]
+                                                    (ws-session/send! ws (byte-array (reverse as-vec)))
+                                                    (swap! server-binary-messages conj as-vec)))
+                                    :on-error   (fn [ws error]) ;; TODO - Add test for on-error behaviour
+                                    :on-close   (fn [ws code reason] (swap! connected dec)
+                                                  (deliver closed true))}]
+        (add-websocket-handler handlers path)
+        (let [socket (ws-client/connect (str "ws://localhost:8080" path)
+                                        :on-receive (fn [text] (swap! client-messages conj text))
+                                        :on-binary  (fn [bytes offset len]
+                                                      (let [as-vec (vec bytes)]
+                                                        (swap! client-binary-messages conj as-vec)
+                                                        (deliver binary-client-message true))))]
+          (ws-client/send-msg socket "Hello websocket handler")
+          (ws-client/send-msg socket "You look dandy")
+          (ws-client/send-msg socket (byte-array [2 1 2 3 3]))
+          (deref binary-client-message)
+          (is (= @connected 1))
+          (ws-client/close socket)
+          (deref closed)
+          (is (= @connected 0))
+          (is (= @server-binary-messages [[2 1 2 3 3]]))
+          (is (= @client-binary-messages [[3 3 2 1 2]]))
+          (is (= @client-messages ["Hello client!"
+                                   "You said: Hello websocket handler"
+                                   "You said: You look dandy"]))
+          (is (= @server-messages ["Hello websocket handler"
+                                   "You look dandy"])))))))
+
 (deftest war-test
   (testing "WAR support"
     (with-app-with-config app
@@ -213,12 +267,14 @@
             path-servlet             "/foo"
             path-war                 "/bar"
             path-proxy               "/baz"
+            path-websocket           "/quux"
             get-registered-endpoints (partial get-registered-endpoints s)
             add-context-handler      (partial add-context-handler s)
             add-ring-handler         (partial add-ring-handler s)
             add-servlet-handler      (partial add-servlet-handler s)
             add-war-handler          (partial add-war-handler s)
             add-proxy-route          (partial add-proxy-route s)
+            add-websocket-handler    (partial add-websocket-handler s)
             ring-handler             (fn [req] {:status 200 :body "Hi world"})
             body                     "This is a test"
             servlet                  (SimpleServlet. body)
@@ -228,6 +284,7 @@
                                             (.addMapping (into-array [path-servlet]))))
                                         (contextDestroyed [this event]))]
             war                      "helloWorld.war"
+            websocket-handlers       {:on-connect (fn [ws])}
             target                   {:host "0.0.0.0"
                                       :port 9000
                                       :path "/ernie"}
@@ -242,6 +299,7 @@
         (add-war-handler (str dev-resources-dir war) path-war)
         (add-proxy-route target path-proxy)
         (add-proxy-route target2 path-proxy {})
+        (add-websocket-handler websocket-handlers path-websocket)
         (let [endpoints (get-registered-endpoints)]
           (is (= endpoints {"/ernie" [{:type :context :base-path dev-resources-dir
                                        :context-listeners []}]
@@ -255,7 +313,8 @@
                             "/baz" [{:type :proxy :target-host "0.0.0.0" :target-port 9000
                                      :target-path "/ernie"}
                                     {:type :proxy :target-host "localhost" :target-port 10000
-                                     :target-path "/kermit"}]}))))))
+                                     :target-path "/kermit"}]
+                            "/quux" [{:type :websocket}]}))))))
 
   (testing "Log endpoints"
     (with-test-logging
--- a/README.md
+++ b/README.md
@@ -55,6 +55,7 @@ This is the protocol for the current implementation of the `:WebserverService`:
 (defprotocol WebserverService
   (add-context-handler [this base-path context-path] [this base-path context-path options])
   (add-ring-handler [this handler path] [this handler path options])
+  (add-websocket-handler [this handlers path] [this handler path options])
   (add-servlet-handler [this servlet path] [this servlet path options])
   (add-war-handler [this war path] [this war path options])
   (add-proxy-route [this target path] [this target path options])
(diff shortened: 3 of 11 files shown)


cprice404:
@richardc thanks a bunch for doing this.  I have two general comments before I get into a deep-dive review:

1. You did way more work than I did on that first commit; I'm not sure the attribution to me is really fair to you :)  You should change it if you end up rebasing or tweaking anything.
2. I'd like to try to balance our somewhat-conflicting interests of "get this merged and released ASAP so that other projects can build off of it" and "make sure we're 100% happy with the API long-term", so I'm thinking maybe we should try to flag some of this stuff as "experimental", just to give us the flexibility to tweak API a bit before we commit ourselves to it via semver.

Not sure exactly how item 2 manifests itself.  Maybe we add another package level called 'experimental' to the new websocket-related namespaces?  Maybe we log a warning message that says 'experimental' when you call some of the entry-point functions, like add-ws-handler?

What do you think?

Will begin a more thorough review momentarily.

cprice404:
@camlow325 just a heads up that this is probably coming in before too long.

src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
```diff
@@ -24,6 +24,9 @@
 (def ContextHandlerOptions
   (dissoc (merge jetty9-core/ContextHandlerOptions RouteOption) :server-id))
 
+(def WebsocketHandlerOptions
+  (dissoc (merge jetty9-core/CommonOptions RouteOption) :server-id))
```
cprice404: Just a sanity check: so, so far, we don't have any new options that are specific to a websocket handler?

It looks like we don't have a schema for `RingHandlerOptions`, so I'm wondering if we need a schema for this one?  At least until we have some new options that are specific to it?  (it looks like `add-ring-handler` just uses `RouteOption` directly?)

I don't have a strong opinion about this, just taking a survey of the existing code.  @fpringvaldsen do you have any thoughts?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_websockets.clj
```diff
@@ -0,0 +1,105 @@
...
+  (ssl? [this]
+    (.. this (getSession) (getUpgradeRequest) (isSecure)))
+  (peer-certs [this]
+    (.. this (getCerts)))
+  (idle-timeout! [this ms]
+    (.. this (getSession) (setIdleTimeout ^long ms)))
+  (connected? [this]
+    (. this (isConnected))))
+
+(defn- do-nothing [& args])
```
cprice404: Would it maybe be worth extending this so that it logged a message, at debug or trace level, indicating that an event had been handled by the default no-op handler?  Maybe change it to:

```clj
(defn- do-nothing [event & args]
   (log/debugf "No handler defined for websocket event '%s' with args: '%s'" event args))
```

and then close over it with `(partial do-nothing :on-connect)`, `(partial do-nothing :on-error)`, etc.?

Not 100% sure whether this is necessary, just thinking that maybe it might be useful for debugging purposes?

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
```diff
@@ -171,6 +173,41 @@
...
+        (add-websocket-handler handlers path)
+        (let [socket (ws-client/connect (str "ws://localhost:8080" path)
+                                        :on-receive (fn [text] (swap! client-messages conj text)))]
+          (ws-client/send-msg socket "Hello websocket handler")
+          (ws-client/send-msg socket "You look dandy")
+          ;; Websockets by their nature are a chat protocol so we have
+          ;; to wait a small amount of time for the responses sent by
+          ;; the :on-text handler to make it back before we can
+          ;; disconnect
+          (Thread/sleep 200)
```
cprice404: I think maybe you could implement this by creating a `(promise)` variable in your `let` block, and then, in the `on-text` handler, after your `(swap! server-messages ...)` call, just do `(deliver mypromise true)`.  Then instead of the sleep here you could just dereference the promise with `@mypromise`, and that'd make sure that you blocked for exactly as long as was necessary with no race conditions.  Would that work?

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
```diff
@@ -171,6 +173,41 @@
...
+            path                  "/test"
+            connected             (atom 0)
+            server-messages       (atom [])
+            client-messages       (atom [])
+            handlers              {:on-connect (fn [ws]
+                                                 (ws-session/send! ws "Hello client!")
+                                                 (swap! connected inc))
+                                   :on-text    (fn [ws text]
+                                                 (ws-session/send! ws (str "You said: " text))
+                                                 (swap! server-messages conj text))}]
```
cprice404: Minor nitpicks w/rt testing:

1. It'd probably be good to have a test that exercised each one of the different types of handlers if at all possible.  If we were able to flag this stuff as `experimental` in an obvious way, as I mentioned in my earlier comment, then I might be satisfied with putting some kind of TODO comment here that says "we need to add tests for `on-close`, ... as soon as we move this out of "experimental".
2. I'm not sure how much testing we do for other types of handlers on the webrouting side of things, but it seems like we might need at least a really basic test for that.  @fpringvaldsen what do you think?

cprice404:
Overall this looks really good to me, thanks @richardc !  I would like to see if we can get @fpringvaldsen to review it, and give @camlow325 an opportunity to do so if he has time / interest.  Other than that and the handful of minor comments I left, I'm +1, and if we were able to flag this as 'experimental' I'd have no major concerns about getting this merged and doing a release very soon.

src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
richardc: On reading this again I agree that `RouteOption` would fit fine, I was just copying the pattern that seemed apparent from `add-servlet-hanlder`, but as we have no specific options then sure, just passing RouteOption would make sense.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
richardc: It would if we simplify this down to *one* message pair, but as it's two messages we need two futures?  An map of futures?

If just sending one pair of messages works then swapping this for a future is fine.

richardc:
@cprice404 1) yeah that attribution is because I repeatedly squashed into your initial spike 20ee82b294e6048f80582d6ea6c536d72b5647f2.  When squashing the next set of fixups I'll re-assign authorship too.

2) so `puppetlabs.experimental.websockets.client`? Seems fine to me

src/puppetlabs/trapperkeeper/services/webserver/jetty9_websockets.clj
richardc: I have no strong feeling on if we need it or not, but I'm happy enough to add it.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
richardc: Adding tests for all the handlers (apart from maybe `on-error`) should be easy enough, so we'll do it.

src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
fpringvaldsen: I think I'm also in favor of using RouteOption since that's what `add-ring-handler` is doing.

fpringvaldsen: Actually...I think we should change the name of schema to `CommonOptions` and use it for the ring handler as well. There is one option, `:redirect-if-no-trailing-slash`, that applies to all handlers. That option is not currently supported for `add-ring-handler` in the webrouting service since it's currently using `RouteOption`, which only supports the `:route-id` option, whereas this schema will also allow the `:redirect-if-no-trailing-slash` option in the webrouting service.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_websockets.clj
fpringvaldsen: :+1: on adding the log message

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
fpringvaldsen: @cprice404 @richardc I agree, even if this feature is experimental I think it would be best to have a basic test for all the handler types to ensure everything is working properly.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
```diff
@@ -301,4 +343,4 @@
                                                                 :follow-redirects false})]
           (is (= (:status response) 302))
           (is (= (get-in response [:headers "location"]) "http://localhost:8080/hello/"))
-          (is (= (get-in response [:opts :url]) "http://localhost:8080/hello")))))))
\ No newline at end of file
+          (is (= (get-in response [:opts :url]) "http://localhost:8080/hello")))))))
```
fpringvaldsen: Nitpick, but I think we should keep the newline at the end of this file.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
fpringvaldsen: From looking at this test it seems to me like one message would probably be fine, in which case we would just need two future, one for `server-messages` and one for `client-messages`

src/puppetlabs/websockets/client.clj
```diff
@@ -0,0 +1,10 @@
+(ns puppetlabs.websockets.client)
+
+(defprotocol WebSocketProtocol
```
fpringvaldsen: I think it would be nice to add some docstrings to this protocol and its functions to explain what they are/what they do, and also what this protocol is being used for in terms of TK Jetty9 implementation details.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_websockets.clj
```diff
@@ -0,0 +1,105 @@
...
+        (let [^WebSocketAdapter this this]
+          (proxy-super onWebSocketConnect session))
+        (on-connect this))
+      (onWebSocketError [^Throwable e]
+        (on-error this e))
+      (onWebSocketText [^String message]
+        (on-text this message))
+      (onWebSocketClose [statusCode ^String reason]
+        (let [^WebSocketAdapter this this]
+          (proxy-super onWebSocketClose statusCode reason))
```
fpringvaldsen: Do we need to do this for `onWebSocketError`, `onWebSocketText`, and `onWebSocketBinary` as well, or do we want to ignore the `WebSocketAdapter` implementation of those?

fpringvaldsen:
@richardc I had a few comments/questions, but nothing major.

@cprice404 I'm :+1: on you merging this once my feedback is addressed and you're satisfied with the state of this PR.

cprice404:
> 2) so puppetlabs.experimental.websockets.client? Seems fine to me

Yeah, that works for me.  Also maybe `puppetlabs/trapperkeeper/services/webserver/jetty9_websockets.clj` -> `puppetlabs/trapperkeeper/services/webserver/experimental/jetty9_websockets.clj` ?

src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
cprice404: @fpringvaldsen there's already a `CommonOptions`, isn't there?

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
cprice404: My order of preference would be:

1. Figure out a way to write the test with two messages and no race conditions (maybe using a Queue structure from java.util.concurrent?), but I grant, possibly diminishing returns here.
2. Simplify to a single message pair and use a promise (and I think I'm pretty much fine with this option, especially if we're going to flag this as experimental for now)
3. Thread/sleep / race conditions :(

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
cprice404: @richardc awesome.  The error one is the one I was thinking might be tricky, too, so feel free to timebox that and then just leave a TODO if you can't figure out a sane way to test it (again, assuming we're flagging this as experimental).

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
richardc: You're reading it backwards.  Original file *lacked* a trailing newline.  My editor on save ensures there's a trailing newline, so we get this whitespace change.

If need be I'll remove this new trailing newline.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
richardc: I realised I can restructure this to just have the promise delivered in the `on-close` handler, so as long as that `on-close` event is correctly dispatched we should be fine.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
fpringvaldsen: Ah, sorry about that. I think we should definitely have a trailing newline at the end of files, so no changes necessary here

src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
fpringvaldsen: @cprice404 There is in the `jetty9-core` namespace, but the set of common options for the webserver service and the webrouting service are different, since the webrouting service also needs to support the `:route-id` option in addition to the `:redirect-if-no-trailing-slash` option.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_websockets.clj
richardc: The `WebsocketAdapter` implementation of those methods are empty bodies:  https://github.com/eclipse/jetty.project/blob/master/jetty-websocket/websocket-api/src/main/java/org/eclipse/jetty/websocket/api/WebSocketAdapter.java#L54-L80  so for consistency we could dispatch to them, but it seems needless.

src/puppetlabs/websockets/client.clj
richardc: Will do.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_websockets.clj
fpringvaldsen: Hmm, I could go either way on this.

I was thinking it would be beneficial to call into them in case future versions of Jetty change the `WebsocketAdapter` implementation of those methods and we ever want to upgrade. We ran into an issue similar to that a while ago, wherein we were overriding a method in Jetty's `ProxyServlet` class with `proxy`, but the implementation of that method changed in a later version of Jetty. This was causing a bunch of issues with our tests when we tried to upgrade, and it took us a long time to figure out that the Jetty-side implementation of the method we were overriding had changed, and that we had to change our implementation of the method as well. I think @camlow325 did that work, so he might have some more insight.

Saying that, while I have very little knowledge of Websockets, my guess is that the implementation of those methods is unlikely to change, in which case I think it would be fine to not call into them.

I think I'd lean towards calling into them for the extra safety, but I don't have a super strong opinion either way. @richardc @cprice404 what do you think?

cprice404: I'm in favor of calling the parent methods to try to give us more forward-compatibility with future versions of Jetty.

richardc: :+1:

richardc:
@cprice404 @fpringvaldsen have reworked in accordance to our discussions and re-pushed.

src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
```diff
@@ -123,6 +123,14 @@
         add-servlet-handler (:add-servlet-handler webserver-service)]
     (add-servlet-handler servlet path opts)))
 
+(schema/defn ^:always-validate add-websocket-handler!
+  [context webserver-service
+   svc :- (schema/protocol tk-services/Service)
+   handlers options :- RouteOption]
```
fpringvaldsen: I think the options schema needs to stay as what you had before (merging `RouteOption` with the `CommonOptions` from the Jetty9 core namespace, and dissociating the `:server-id` key), as there is an option in the `CommonOptions` in the Jetty9 core namespace that all handlers need to support (`:redirect-if-no-trailing-slash`). The schema can probably just be called `CommonOptions`, and should be used for `add-ring-handler!` as well in place of `RouteOption`.

fpringvaldsen:
:+1: aside from fixing up the use of the `RouteOption` schema.

src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
richardc: So if I'm interpreting this correctly we want https://github.com/richardc/trapperkeeper-webserver-jetty9/commit/6f7617add3a84574d856a5a6b4aa21034668d539 ?

cprice404:
+1 from me.  @fpringvaldsen feel free to merge when you're happy with the schema stuff.

src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
cprice404: @richardc I think that is what @fpringvaldsen is suggesting, yes.  I'd recommend going ahead and pushing that change up.

@fpringvaldsen if what @richardc pushes up isn't quite what you were thinking, maybe you could go ahead and add one more commit to take it to the finish line and then merge it in?  That way we won't have to wait through another 24-hour cycle to get this one wrapped up.

richardc:
Pushed up the changes.

fpringvaldsen:
:shipit:

PR #121, opened 2015-08-14
(MAINT) Upgrade to ssl-utils 0.8.1

nwolfe:
This keeps it in line with the version used in puppet-server

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -13,7 +13,6 @@
   :pedantic? :abort
   :dependencies [[org.clojure/clojure "1.6.0"]
                  [org.clojure/tools.logging "0.2.6"]
-                 [clj-time "0.5.1"]
                  [prismatic/schema "0.4.0"]
                  [prismatic/plumbing "0.4.2"]
                  [puppetlabs/kitchensink ~ks-version]
--- a/project.clj
+++ b/project.clj
@@ -15,9 +15,11 @@
                  [org.clojure/tools.logging "0.2.6"]
                  [prismatic/schema "0.4.0"]
                  [prismatic/plumbing "0.4.2"]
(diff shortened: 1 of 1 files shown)


cprice404:
+1 but @nwolfe do you have any idea what's going on w/Travis here?  I can't seem to get it to even show me the build history.

nwolfe:
@cprice404 Ahh, pedantic errors regarding clj-time. This project depends on `0.5.1` but ssl-utils brings in `0.7.0`.

I think we should be good to upgrade this clj-time dependency. PR update coming...

nwolfe:
@cprice404 @camlow325 Removed erroneous `clj-time` dependency; unit tests pass and there are no references to it in the code.

nwolfe:
@cprice404 @camlow325 Blarrrrghhh....turns out we had that errant clj-time dependency as a way of dealing with the fact that both kitchensink and trapperkeeper pull in the (old) clj-time dependency as well.

I basically added this reconciliation back in, but using exclusions instead.

Happy to go back to the older style, though I think this is more clear as to what's going on.

Looks like Travis is happy now.

KevinCorcoran:
LGTM

camlow325:
I'm okay with this, although I think I'd like to have @cprice404 take a look at this before merging it since I expect that he may have some strong opinions on how we do the dependencies - like maybe a preference for continuing to express the `clj-time` version explicitly here.

cprice404:
We need to talk about the exclusion vs. not-exclusion thing as a team.  I'll try to remember to pull that to the top in the next festivus or something.  Until then I'm not going to issue any edicts about it, so, I guess I'll go ahead and merge this so it can get into the upcoming release.

PR #122, opened 2015-08-18
(maint) add variant of websockets/close! with reason

richardc:
Here we extend the WebSocketProtocol to add a three-arg variant of
close! which allows for the reason we are closing a connection to
be specified.


cprice404:
+1 and would prefer to get this in before we do our initial release with websocket support so that we don't have to worry about changing the API later.

PR #124, opened 2015-08-21
TK-270 Restore and test for 1-arity version of websockets/close!

richardc:
Here we correct the implementation of the WebsocketProtocol for
WebSocketSend to not stomp the arity-1 version of close!.


cprice404:
@richardc ping me tomorrow and we can do a z release

PR #126, opened 2015-08-27
(maint) update the changelog for 1.4.1

richardc:


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 1.4.1
+
+This is a maintenance release.
```
cprice404: I think we should probably say "bugfix release"?

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+
+* [TK-270](https://tickets.puppetlabs.com/TK-270) Restore the definition of
```
cprice404: I would maybe say "fix a bug that prevented the use of the 1-arity ..."; not too picky on the wording other than I'd probably be explicit in calling out that it was a bugfix.

CHANGELOG.md
richardc: :+1:

PR #127, opened 2015-10-02
Add get-server to WebroutingService protocol

ploubser:
Here we add a get-server function that, similar to get-route, will return
the configured server for the current service with a specified id.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service.clj
+++ b/src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service.clj
@@ -9,6 +9,7 @@
 
 (defprotocol WebroutingService
   (get-route [this svc] [this svc route-id])
+  (get-server [this svc] [this svc route-id])
   (add-context-handler [this svc context-path] [this svc context-path options])
   (add-ring-handler [this svc handler] [this svc handler options])
   (add-servlet-handler [this svc servlet] [this svc servlet options])
@@ -37,6 +38,12 @@
 
   (get-route [this svc route-id]
              (core/get-route (service-context this) svc route-id))
+
+  (get-server [this svc]
+              (core/get-server (service-context this) svc nil))
+
+  (get-server [this svc route-id]
+              (core/get-server (service-context this) svc route-id))
 
   (add-context-handler [this svc base-path]
                     (core/add-context-handler! (service-context this)
--- a/src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
@@ -101,6 +101,16 @@
                                                                  route-id)]
     (:route endpoint-and-server)))
 
+(schema/defn ^:always-validate get-server :- (schema/maybe schema/Str)
+  [context
+   svc :- (schema/protocol tk-services/Service)
+   route-id]
+  (let [svc-id              (keyword (tk-services/service-symbol svc))
+        endpoint-and-server (get-endpoint-and-server-from-config context
+                                                                 svc-id
+                                                                 route-id)]
+    (:server endpoint-and-server)))
+
 (schema/defn ^:always-validate add-context-handler!
   [context webserver-service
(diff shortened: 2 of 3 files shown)


ploubser:
@richardc @cprice404

cprice404:
@fpringvaldsen this is basically exactly what you were describing in TK-284, right?

src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service.clj
```diff
@@ -38,6 +39,9 @@
   (get-route [this svc route-id]
              (core/get-route (service-context this) svc route-id))
 
+  (get-server [this svc route-id]
+             (core/get-server (service-context this) svc route-id))
+
```
cprice404: In the protocol you define two signatures, but here you only implement one of them, right?

src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
```diff
@@ -101,6 +101,14 @@
                                                                  route-id)]
     (:route endpoint-and-server)))
 
+(defn get-server
+[context svc route-id]
+(let [svc-id              (keyword (tk-services/service-symbol svc))
+      endpoint-and-server (get-endpoint-and-server-from-config context
+                                                               svc-id
+                                                               route-id)]
+  (:server endpoint-and-server)))
```
cprice404: I wouldn't mind seeing this fn be a `schema/defn`.

cprice404:
Needs some tests but I'm +1 on the idea in general, and, this same thing hit us just the other day so we would be happy to help get this pushed through.

ploubser:
@cprice404 Excellent. In that case I'll rework and turn into a mergable PR.

camlow325:
@cprice404 Not that I'm opposed to this PR, but I think the motivation for TK-284, IIUC, is to be able to get the "configuration" for the server by a service / route-id as opposed to the actual Jetty `Server` object, which is what I think this PR does instead.  I believe TK-284 was motivated by the override webserver configuration stuff that we do in Puppet Server.  I think the idea of having a "get configuration" capability that you can use independently of the actual server object which gets constructed - as a separate PR.

richardc:
If I remember correctly, this isn't getting the server object, it's getting the server key from here https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/1.4.1/src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj#L39   which is either going to be nil or a string with the name of a webserver (or an exception if it's not mounted).

camlow325:
@richardc Ah, yes, you're right.  Thanks.  So, nevermind then.  I think this would cover what TK-284 was describing then.

richardc:
Well TK-284 is probably more a `(get-webserver-config route-id)` but this is a compliments the existing `(get-route route-id)` call, so I'm not sure which is ultimately preferable.

camlow325:
Yeah, maybe that or callers could use the combination of the result from this new `(get-server)` call with a complementary `(get-webserver-config server)` - so that the caller wouldn't otherwise have to do the latter by accessing the `webserver` map from the TK context directly?

ploubser:
@cprice404 Updated PR.

src/puppetlabs/trapperkeeper/services/webrouting/webrouting_service_core.clj
```diff
@@ -101,6 +101,16 @@
                                                                  route-id)]
     (:route endpoint-and-server)))
 
+(schema/defn ^:always-validate get-server
```
cprice404: can you add schema for the return type:  Just glancing at the code it's hard to tell whether it's a server id or an actual Jetty `Server` object.

cprice404:
This looks fine to me other than the one request about the schema for the return type.  That said, @fpringvaldsen has the most context on this topic lately so I'd love to get his thoughts.  (He's on PTO today but back tomorrow.)

ploubser:
Updated and thanks!

fpringvaldsen:
:+1:

PR #128, opened 2015-10-07
(maint) Prep for 1.5.0 release

ploubser:


doc/webrouting-service.md
```diff
@@ -78,6 +79,18 @@ service with the id you specify.
 Note that the one argument version cannot be used with a service that
 has multiple webroutes configured.
 
+#### `get-server`
+
+This function allows you to get the server for a particular service
+as configured in your configuration file. The one-argument version will
+return the server configured for the current service in a single-route
+configuration. The two
+argument version will return the server configured for the current
```
fpringvaldsen: I think this should mention that the id you specify is for a route configured for the current service, and that the two-argument version will retrieve the server configured for that particular route.

fpringvaldsen:
@ploubser Thanks! I think we should also bump the version in `project.clj` to `1.5.0-SNAPSHOT` so that the Jenkins release job will release the correct version. Other than that and the one minor piece of feedback above, I'm :+1:

PR #129, opened 2016-02-13
(TK-301) fix hup memleaks

cprice404:
This first commit in this PR upgrades us to TK 1.3.0, which provides HUP support.  The second commit fixes a memory leak relating to the JMX MBean registry, adds test coverage around this leak, and updates all of the rest of the tests to ensure that they shut down any server objects that they create cleanly so that we don't pollute the other test namespaces with leftover state.


runtests.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
```
rlinehan: was this supposed to be committed?

cprice404: nope :)

test/clj/puppetlabs/trapperkeeper/testutils/webserver.clj
```diff
@@ -55,3 +58,25 @@
      ~port-var
      {}
      ~@body))
+
+(defn get-jetty-mbean-object-names
+  "TODO SPLAIN"
```
rlinehan: Docstring still todo?

test/clj/puppetlabs/trapperkeeper/testutils/webserver.clj
```diff
@@ -55,3 +58,25 @@
...
+                 (ObjectName. "org.eclipse.jetty.jmx:type=mbeancontainer,*"))))
+
+(defn assert-clean-shutdown
+  "TODO SPLAIN"
+  [f]
+  (f)
+  ;; This sucks, because if this assertion fails, it will not give any clue as to
+  ;; which test caused it to fail beyond just the test namespace.  However, I tried
+  ;; several things (such as throwing an exception here rather than having an assertion),
+  ;; and nothing gave any better debugging info because the metadata about which
```
rlinehan: Docstring still todo?

test/clj/puppetlabs/trapperkeeper/testutils/webserver.clj
```diff
@@ -55,3 +58,25 @@
...
+(defn get-jetty-mbean-object-names
+  "TODO SPLAIN"
+  []
+  (let [mbeanserver (ManagementFactory/getPlatformMBeanServer)]
+    (.queryNames mbeanserver
+                 nil
+                 (ObjectName. "org.eclipse.jetty.jmx:type=mbeancontainer,*"))))
+
+(defn assert-clean-shutdown
+  "TODO SPLAIN"
```
rlinehan: Docstring still todo?

test/clj/puppetlabs/trapperkeeper/testutils/webserver.clj
```diff
@@ -55,3 +58,25 @@
...
+  [f]
+  (f)
+  ;; This sucks, because if this assertion fails, it will not give any clue as to
+  ;; which test caused it to fail beyond just the test namespace.  However, I tried
+  ;; several things (such as throwing an exception here rather than having an assertion),
+  ;; and nothing gave any better debugging info because the metadata about which
+  ;; test we're running and the call stack for the test function itself are simply
+  ;; not available from inside a fixture.  So I decided that even those these aren't
+  ;; super fun to debug, it was better than any other option in terms of enforcing
+  ;; that the tests clean up after themselves properly.
```
rlinehan: I noticed that everywhere this is used, it is used as with `use-fixtures :each`. Since this won't provide any information beyond what test namespace the issue is in, it seems like it would provide the same amount of information if it was used with `use-fixtures :once` and perhaps not have as much overhead?

test/clj/puppetlabs/trapperkeeper/testutils/webserver.clj
```diff
@@ -55,3 +58,25 @@
...
      {}
      ~@body))
+
+(defn get-jetty-mbean-object-names
+  "TODO SPLAIN"
+  []
+  (let [mbeanserver (ManagementFactory/getPlatformMBeanServer)]
+    (.queryNames mbeanserver
+                 nil
+                 (ObjectName. "org.eclipse.jetty.jmx:type=mbeancontainer,*"))))
```
rlinehan: I noticed that in https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj#L116 we're using `clojure.java.jmx/mbean-names`, which looks like it does almost exactly this. Would it make sense to use that here too?

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
```diff
@@ -207,9 +205,13 @@
       (update-in [connector-keyword :so-linger-milliseconds] identity)
       (update-in [connector-keyword :idle-timeout-milliseconds] identity)))
 
+
+(def default-http-connector-config
+  {:jmx-enable "false"})
```
rlinehan: Why was this (and the change on line 214) necessary? (If I change 214 back to just `config` everything still seems to work.) Also, it seems kind of odd to call this `default-http-connector-config` if the default is set to true in our code... but I feel like I am probably missing what exactly this is used for?

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
```diff
@@ -444,15 +446,23 @@
...
-                (merge base-config
-                       {:post-config-script (str "Object x = null; x.toString();")}))))))))
+              (let [context (jetty9/initialize-context)]
+                (try
+                  (jetty9/start-webserver!
+                   context
+                   (merge base-config
+                          {:post-config-script (str "Object x = null; x.toString();")}))
+                  (finally
+                    (jetty9/shutdown context))))))))))
```
rlinehan: There's a couple new wolfe principle violations here and from the `testing` block above this one, probably due to the tk logging update.

rlinehan: I think there's also one from https://github.com/cprice404/trapperkeeper-webserver-jetty9/blob/bug/master/TK-301-fix-HUP-memleaks/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj#L760 but I can't comment on that in the diff of this PR easily.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -334,10 +341,12 @@
                  (Server. pool)
                  (Server.))]
     (when (:jmx-enable config)
+      ;(log/error (IllegalStateException. "CREATING SERVER WITH JMX ENABLED"))
```
rlinehan: Looks like this is a stray debug log statement.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -547,6 +560,9 @@
 
 (schema/defn ^:always-validate shutdown
   [webserver-context :- ServerContext]
+  (when-let [{:keys [mbean-container]} @(:state webserver-context)]
+    (log/debug "Cleaning up JMX MBean container")
+    (.destroy mbean-container))
```
rlinehan: This seems to be throwing a null pointer exception with `jmx-enabled: "false"` set. It looks like if a key is present in a map, even if the value of that key is falsy, `when-let` will use it. (actually, from reading up on this a bit more, it looks like the `when-let` is completely dependent on the value on the right for the conditional).

Seems like it might be a good idea to have a test of shutting down the server when jmx reporting is disabled too?

test/clj/puppetlabs/trapperkeeper/testutils/webserver.clj
cprice404: Oh, I didn't realize we'd already introduced a dependency on that clojure lib.  Yes, we should definitely use it.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
cprice404: Good call; I think maybe I just added that in there when I was trying to find tests that weren't cleaning up properly and were leaving behind polluted state for other tests, but then maybe it turned out that it wasn't actually this namespace that was the culprit, so that change isn't necessary.  Will revert.

cprice404:
@rlinehan I believe I've addressed all of your comments.  Also rebased to remove the commit where we were adding the ring example to the uberjar for testing.

rlinehan:
Travis failed with `*** buffer overflow detected ***: java terminated` which is slightly scary. I'll re-run.

camlow325:
@cprice404, @rlinehan - I wonder if we need to apply that "travis buffer overflow workaround" like we did for other repos - see https://github.com/puppetlabs/trapperkeeper/blob/1.3.0/.travis.yml#L14-L19.

rlinehan:
I'm :+1:  on this. I put up a PR for the buffer overflow issue at https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/130.

cprice404:
@camlow325 we are going to need to get this merged and released and use it in OSS before we ship, so if you are interested in reviewing, let me know when you think you might get to it.  @rlinehan is going to be demo'ing what we saw in MAT/YourKit this afternoon so if you want to wait until you've heard us talk through that, that's fine... but this change seems necessary even w/o the memory testing data.

camlow325:
@cprice404 I'll take a look this morning.  Don't think I need to wait for the demo.

dev-resources/logback.xml
```diff
@@ -5,11 +5,10 @@
         </encoder>
     </appender>
 
-    <logger name="puppetlabs" level="error"/>
-    <logger name="log4j.logger.org.eclipse.jetty.server" level="warn"/>
-    <logger name="log4j.logger.org.eclipse.jetty.util.log" level="warn"/>
+    <logger name="org.apache.http" level="error"/>
```
camlow325: Just curious, were these changes just useful for dev purposes or were they necessary to fix some of the "logging tests" in conjunction with the corresponding bumps to the new logutils that came with the bump to the new tk version?

cprice404: the latter

runtests.sh
camlow325: Still planning on removing this from the PR?

cprice404: ugh, yes

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -133,6 +128,18 @@
...
 
+(def ServerContextState
+  {:mbean-container (schema/maybe MBeanContainer)
+   :overrides-read-by-webserver schema/Bool
+   :overrides (schema/maybe {schema/Keyword schema/Any})
+   :endpoints RegisteredEndpoints
+   :ssl-context-factory (schema/maybe SslContextFactory)})
+
+(def ServerContext
+  {:state     (schema/atom ServerContextState)
```
camlow325: Yay, `schema/atom`!

cprice404:
@camlow325 I just added a commit that removes that errant file, plus rebased on latest master in hopes of picking up @rlinehan 's change to fix the buffer overflow error on travis.

test/clj/puppetlabs/trapperkeeper/testutils/webserver.clj
camlow325: If we did it as `:once`, though, is it possible that one or more bad tests in the namespace don't clean themselves up but if the last one run does we might not assert at the end?  If so, I think `:each` might still be better.  Hopefully the call to get the `mbean-names` would be cheap enough that we're not adding significant performance overhead to each test by doing this?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -526,7 +534,11 @@
   []
   (let [^ContextHandlerCollection chc (ContextHandlerCollection.)]
     {:handlers chc
-     :state (atom {:endpoints {}})
+     :state (atom {:endpoints {}
+                   :mbean-container nil
+                   :overrides-read-by-webserver false
```
camlow325: Thanks for doing this.  Feeling very ashamed now that I didn't do this in the first place.... :disappointed:

test/clj/puppetlabs/trapperkeeper/testutils/webserver.clj
cprice404: The mbean registry cleanup is a per-Jetty-server-instance thing, so if even a single test doesn't clean up, the assertion will fail at the end up of the namespace.  i.e. if one test doesn't clean up after its `Server` instance, a subsequent test that cleaned up after itself wouldn't clean up for the earlier test.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
cprice404: I don't think `atom` even existed back then.  Besides, we were young, and needed the money.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
```diff
@@ -444,15 +447,25 @@
...
               IllegalArgumentException
               #"Invalid script string in webserver 'post-config-script' configuration"
-              (jetty9/start-webserver!
-                (jetty9/initialize-context)
-                (merge base-config
-                       {:post-config-script (str "AHAHHHGHAHAHAHEASD!  OMG!")})))))
+              (let [context (jetty9/initialize-context)]
+                (with-test-logging
+                 (try
+                   (jetty9/start-webserver!
```
camlow325: Maybe we should look into creating a `with-webserver-running` testutils macro to cut down on the need to have a try/finally and shutdown call everywhere these get used?  Was thinking we might have done something like that already at some point but maybe not?  Just a suggestion - wouldn't hold up the PR for this...

test/clj/puppetlabs/trapperkeeper/testutils/webserver.clj
camlow325: Ok, that makes sense.  I'm good with `:once` per namespace, then.

camlow325:
:+1: and looks like Travis passed on all OSes with no "most-heinous" Wolfe Principle violations to be found.  So merging...

PR #134, opened 2016-03-08
(TK-338) Catch and log error for TimeoutException during stop

cprice404:
When a TK app is sent a HUP signal to request a restart, Jetty will
attempt to finish off any active requests gracefully before shutting
down (and subsequently restarting).  However, there is a setting
called `stopTimeout` that determines how long the server will wait
to "gracefully" shut down, before it gets more aggressive in killing
off threads.

Prior to this commit, if we were in the midst of a HUP restart, and
this timeout was exceeded during the graceful shutdown attempt,
Jetty would throw a `TimeoutException` that was treated as fatal
by TK.  This commit catches that exception and logs an error message,
but allows the server to continue on and restart Jetty.

I've done a decent amount of manual testing on this (including an
overnight run that included a HUP every 10 minutes, and some
analysis of heap dumps taken at a few places during that run), and
it seems to behave well.

I need to add tests for this, but in the interest of time, I'm
currently planning to do that in a separate commit / PR.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
@@ -558,13 +558,19 @@
     (merge options overrides)))
 
 (schema/defn ^:always-validate shutdown
-  [webserver-context :- ServerContext]
+  [{:keys [server] :as webserver-context} :- ServerContext]
   (when-let [mbean-container (:mbean-container @(:state webserver-context))]
     (log/debug "Cleaning up JMX MBean container")
     (.destroy mbean-container))
   (when (started? webserver-context)
     (log/info "Shutting down web server.")
-    (.stop (:server webserver-context))
+    (try
+      (.stop server)
+      (catch TimeoutException e
+        (log/errorf e
+                    (str "Web server failed to shut down gracefully in configured "
+                         "timeout period (%s); cancelling remaining requests.")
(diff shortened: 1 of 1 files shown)


src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -558,13 +558,19 @@
...
     (.destroy mbean-container))
   (when (started? webserver-context)
     (log/info "Shutting down web server.")
-    (.stop (:server webserver-context))
+    (try
+      (.stop server)
+      (catch TimeoutException e
+        (log/errorf e
+                    (str "Web server failed to shut down gracefully in configured "
+                         "timeout period (%s); cancelling remaining requests.")
```
camlow325: Nitpick - since the underlying type is a `Long`, would it be better for the format string to have a `%d` instead of a `%s`?  Doesn't matter to me either way, just curious what you think.

camlow325:
:+1:

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
cprice404: meh, seems like it works ok this way

rlinehan:
:+1:

rlinehan:
tests passed locally, not gonna bother waiting for travis.

PR #137, opened 2016-03-10
(TK-339) add test for app restarting if stop timeout expires

cprice404:


test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -84,38 +96,28 @@
    (validate-ring-handler base-url config http-get-options :default))
   ([base-url config http-get-options server-id]
     (with-app-with-config app
-      [jetty9-service]
+      [jetty9-service
+       hello-webservice]
       config
-      (let [s                   (tk-app/get-service app :WebserverService)
```
fpringvaldsen: copypasta everywhere

fpringvaldsen:
:+1: 

Travis failed, but it seemed like it might be transient, so I restarted the build

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -84,38 +96,28 @@
...
-        (add-ring-handler ring-handler path {:server-id server-id})
-        (let [response (http-get
-                         (format "%s%s/" base-url path)
-                         http-get-options)]
-          (is (= (:status response) 200))
-          (is (= (:body response) body)))))))
+      (let [response (http-get
+                      (format "%s%s/" base-url hello-path)
+                      http-get-options)]
+        (is (= (:status response) 200))
```
camlow325: I guess in our next round of "make the tests less stupid" we should pull these assertions directly into the tests rather than doing them in a helper function?  Don't think we need to worry about that for this PR...

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -729,7 +712,34 @@
...
+
+  (testing "tk app can still restart even if stop timeout expires"
+    (let [in-request-handler (promise)
+          sleepy-service (tk-core/service
+                          [[:WebserverService add-ring-handler]]
+                          (init [this context]
+                                (add-ring-handler
+                                 (fn [_]
+                                   (deliver in-request-handler true)
+                                   (Thread/sleep 2000)
```
camlow325: Rather than just putting a sleep here, might it be better to just have a line here that waits on a promise that gets delivered from the main thread just after the `tk-app/restart` call occurs?  That way, we wouldn't have the possible race condition where this sleep might finish and the response start to be delivered before Jetty fully gets into its `stop` sequence.

camlow325:
:+1: with just the minor note / question on the sleep timing in the new restart test.

Cleanup of the copypasta on the other tests is .

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
cprice404: That does sound better, I'll take a swing at it when I get a sec.

cprice404: Hmm.  Well, this handler gets called twice, so I would/will need to use an atom wrapping the promise, and use a fresh promise for the second request.

Also, `restart` is synchronous, so I would have to call it on a future in order to be able to deliver the promise after calling the restart... will think about it some more but starting to think this might make the test significantly more complex?

camlow325: We only need the delay to happen for the first `get` request right?  For the second one, I was thinking we wouldn't want a delay anyway - just slows things down.  If that's true, seems like just delivering on the initial promise would be sufficient and the deref done in the handler the second request would effectively be a no-op because it was delivered for the first one already.

Is it a problem for this test if we're waiting in the ring handler function for the blocking `restart` to finish?  The `restart` should still finish, right?  And not throw an exception anymore per the previous code changes we merged?  Might leave the state of the first request in a somewhat ambiguous state, I suppose.  Not sure if it will have been interrupted on waiting for the promise by Jetty sending a thread interrupt during the shutdown?  Either way, seems like the original request will still fail from the client perspective by virtue of the server being torn down during Jetty's stop.

cprice404: Yeah that's true, it's fine for the first request to block, and it seems like it's also fine to wait to deliver the promise until after the restart has completed.

cprice404:
@camlow325 I added a commit that removes the sleep from that test.

camlow325:
:+1:

PR #138, opened 2016-03-14
(TK-133) Safer cleanup of mbeans during stop

cprice404:
This commit modifies our 'stop' logic so that we won't attempt
to destroy/unregister the mbeans more than once, in the event
that 'stop' gets called multiple times by TK.

Unfortunately, the main symptom that we see when we do try to
destroy/unregister the mbeans multiple times is a slew of really
nasty log messages from Jetty, which meant that the test for this
change needed to mess with test logging stuff, which had a ripple
effect that spilled over into some other files.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(def tk-version "1.3.0")
+(def tk-version "1.3.1")
 (def ks-version "1.3.0")
 (def jetty-version "9.2.10.v20150310")
 
@@ -11,18 +11,21 @@
   ;; dependencies. Also supports :warn to simply emit warnings.
   ;; requires lein 2.2.0+.
   :pedantic? :abort
-  :dependencies [[org.clojure/clojure "1.6.0"]
+  :dependencies [[org.clojure/clojure "1.7.0"]
+
+                 ;; begin version conflict resolution dependencies
+                 [org.slf4j/slf4j-api "1.7.7"]
+                 ;; end version conflict resolution dependencies
+
                  [org.clojure/java.jmx "0.2.0"]
                  [org.clojure/tools.logging "0.2.6"]
                  [prismatic/schema "1.0.4"]
-                 [prismatic/plumbing "0.4.2"]
 
-                 ;; Let ssl-utils bring in the clj-time dependency
                  [puppetlabs/ssl-utils "0.8.1"]
-                 [puppetlabs/kitchensink ~ks-version :exclusions [clj-time]]
-                 [puppetlabs/trapperkeeper ~tk-version :exclusions [clj-time]]
+                 [puppetlabs/kitchensink ~ks-version]
+                 [puppetlabs/trapperkeeper ~tk-version]
 
-                 [ch.qos.logback/logback-access "1.1.1"]
+                 [ch.qos.logback/logback-access "1.1.3"]
 
                  [org.codehaus.janino/janino "2.7.8"]
 
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
@@ -561,7 +561,8 @@
   [{:keys [server] :as webserver-context} :- ServerContext]
   (when-let [mbean-container (:mbean-container @(:state webserver-context))]
     (log/debug "Cleaning up JMX MBean container")
-    (.destroy mbean-container))
+    (.destroy mbean-container)
+    (swap! (:state webserver-context) assoc :mbean-container nil))
   (when (started? webserver-context)
     (log/info "Shutting down web server.")
     (try
--- a/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
+++ b/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
@@ -750,22 +750,19 @@
 
 (deftest double-stop-test
   (testing "if the stop lifecycle is called more than once, we handle that gracefully and quietly"
-    (let [log-messages (atom [])]
-      (tk-log-testutils/with-logging-to-atom tk-log/root-logger-name log-messages
-        (let [app (tk-bootstrap/bootstrap-services-with-config
-                   [jetty9-service]
-                   {:webserver {:port 8080}})]
-          (tk-app/stop app)
-          (tk-app/stop app)))
+    (tk-log-testutils/with-logged-event-maps log-events
+      (let [app (tk-bootstrap/bootstrap-services-with-config
+                 [jetty9-service]
+                 {:webserver {:port 8080}})]
+        (tk-app/stop app)
+        (tk-app/stop app))
       ;; we previously had a bug where we could try to unregister mbeans multiple
       ;; times if our tk-j9 stop lifecycle function was called more than once.
       ;; in that case, Jetty would log tons of nasty exceptions at warning level.
       ;; this test just validates that that is not happening.
       (let [log-event-filter #(or (= :warn (:level %))
                                   (= :error (:level %)))
-            mbean-err-logs (->> @log-messages
-                                (map tk-log-testutils/event->map)
-                                (filter log-event-filter))]
+            mbean-err-logs (filter log-event-filter @log-events)]
         (is (= 0 (count mbean-err-logs)))))))
 
 (deftest warn-if-sslv3-supported-test
--- a/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
+++ b/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
@@ -757,12 +757,12 @@
                    {:webserver {:port 8080}})]
           (tk-app/stop app)
           (tk-app/stop app)))
-      ;; if we try to unregister the mbeans more than once Jetty will log a bunch
-      ;; of very loud and terrifying InstanceNotFoundExceptions
-      (let [log-event-filter #(and (= :warn (:level %))
-                                   (not (nil? (:exception %)))
-                                   (re-find #"InstanceNotFoundException"
-                                            (.getClassName (:exception %))))
+      ;; we previously had a bug where we could try to unregister mbeans multiple
+      ;; times if our tk-j9 stop lifecycle function was called more than once.
+      ;; in that case, Jetty would log tons of nasty exceptions at warning level.
+      ;; this test just validates that that is not happening.
+      (let [log-event-filter #(or (= :warn (:level %))
+                                  (= :error (:level %)))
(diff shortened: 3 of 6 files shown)


project.clj
```diff
@@ -1,4 +1,4 @@
-(def tk-version "1.3.0")
+(def tk-version "1.3.1-SNAPSHOT")
```
cprice404: This needs to be fixed; I added a bootstrap testutil to TK that I used in the tests in this PR.

rlinehan: It looks like https://github.com/puppetlabs/trapperkeeper/pull/221 has been merged, so we just need a TK release.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -744,6 +748,26 @@
...
+  (testing "if the stop lifecycle is called more than once, we handle that gracefully and quietly"
+    (let [log-messages (atom [])]
+      (tk-log-testutils/with-logging-to-atom tk-log/root-logger-name log-messages
+        (let [app (tk-bootstrap/bootstrap-services-with-config
+                   [jetty9-service]
+                   {:webserver {:port 8080}})]
+          (tk-app/stop app)
+          (tk-app/stop app)))
+      ;; if we try to unregister the mbeans more than once Jetty will log a bunch
+      ;; of very loud and terrifying InstanceNotFoundExceptions
```
rlinehan: but this is no longer happening with this fix, right? so these exceptions should no longer be logged? it might be good to change the phrasing of this, because right now this comment reads to me like we still expect this to happen.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -744,6 +748,26 @@
...
+      (let [log-event-filter #(and (= :warn (:level %))
+                                   (not (nil? (:exception %)))
+                                   (re-find #"InstanceNotFoundException"
+                                            (.getClassName (:exception %))))
```
rlinehan: yuck

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -744,6 +748,26 @@
              (is (= 200 (:status @response)))
              (is (= hello-body (:body @response))))))))))
 
+(deftest double-stop-test
+  (testing "if the stop lifecycle is called more than once, we handle that gracefully and quietly"
```
rlinehan: Rather than changing the logging for all of the tests, would there be any way to have a logback file just used for this one test (could we start up trapperkeeper in this test with a `{:global logback.xml}` option?)

fpringvaldsen: :+1: to this suggestion, I'd much rather see us change the logging for just this one test

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
rlinehan: I'm guessing that there isn't anything better to test here, given what is stated in the commit message, but this seems really brittle - if anything happened to cause errors here again and if the error changed, we wouldn't have any notification (and, at least I think?, any other errors logged would be hidden in the test). This seems not great. Is there anything else we could test for? Even just that nothing is logged as a warning or an error? Or would that be too general and also brittle in another way?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -561,7 +561,8 @@
   [{:keys [server] :as webserver-context} :- ServerContext]
   (when-let [mbean-container (:mbean-container @(:state webserver-context))]
     (log/debug "Cleaning up JMX MBean container")
-    (.destroy mbean-container))
+    (.destroy mbean-container)
+    (swap! (:state webserver-context) assoc :mbean-container nil))
```
camlow325: If `shutdown` were being called from two different threads at the same time, it seems possible that both threads could have gotten the `:mbean-container` and subsequently try to call `.destroy` on it.  There's probably another way to do this that would involve only calling `.destroy` here if we've successfully moved the current state of the atom from having `:mbean-container` set with the original object to having `:mbean-container` not have that object anymore.  Although if we're eventually going to preclude the possibility of this function being callable from two different threads at the same time, what you already have here is probably sufficient.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -744,6 +748,26 @@
...
+    (let [log-messages (atom [])]
+      (tk-log-testutils/with-logging-to-atom tk-log/root-logger-name log-messages
+        (let [app (tk-bootstrap/bootstrap-services-with-config
+                   [jetty9-service]
+                   {:webserver {:port 8080}})]
+          (tk-app/stop app)
```
camlow325: I understand that the `TrapperkeeperApp` will swalow the exception thrown by the second `stop` call and so looking at the log file might be the only way to determine the benefit of these code changes at more of a "system level".  In this case, though, I'd probably just have the test get the `jetty9-service` instance, call `stop` on it directly and validate that no exception is thrown.  Not as standard of a thing to do from a Trapperkeeper app lifecycle perspective but it seems like it would be more targeted around the specific problem we're trying to validate has been fixed.

camlow325:
I'm basically :+1: on this pending the tk-version bump to a non-SNAPSHOT release.  Left a couple of comments about timing/threading concerns and the design of the test which I don't necessarily consider critical.

project.clj
cprice404: And it probably makes sense to hold off on the TK release until the current slew of tickets is resolved, so this may need to sit a little longer.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
cprice404: fair, i can reword it.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
cprice404: Yeah, it turns out that Jetty is just swallowing the exceptions coming from the JDK MBean API and logging a warning, so there's nothing more concrete to test than log messages.

I think that what you are proposing would be more likely to yield false negatives, whereas mine is more likely to yield false positives.  False positives are probably worse, so I can look into changing it to what you're suggesting.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
cprice404: Are these comments in reference to my changes to `dev-resources/logback.xml`?

In general I agree with you that changing the global logging config for one test is not a good idea.  However, I've come to think that the correct / best practice default config for test logging is to have things set to `warn`.  That way any test where you are explicitly expecting a warning, you can take steps to comply with the Wolfe Principle, but any other test that is unexpectedly logging a warning will show up in the test output rather than being silenced.  This seems like a pretty valuable way to catch unexpected changes, e.g. from upstream libraries, and defaulting the test logs to `error` seems like a bad pattern to me.

So I would actually advocate for that change to `logback.xml` even in the absence of this PR.

Let me know if you disagree with that, or if I've misunderstood these comments entirely :)

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
cprice404: Yeah, that seems like a reasonable suggestion, but I do think that we want to be in a world where it's considered a full-on bug in TK if lifecycle functions are happening on two threads at the same time, and fix that if it comes up.  So I'm inclined to think that it's not worth the extra precautions here?

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
cprice404: That's the thing, though, Jetty wasn't actually throwing an exception.  It was just logging about 100 of them and then swallowing them:

https://github.com/eclipse/jetty.project/blob/93a8d8e05f4120e57bc3c830d5c10e38dd4a9f97/jetty-jmx/src/main/java/org/eclipse/jetty/jmx/MBeanContainer.java#L272

So I really don't think there's anything I can test except logging.

cprice404:
Added a commit that attempts to address @rlinehan 's feedback about the comment in the test, and reduce the likelihood of a false positive.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
camlow325: Yeah, I agree if (when) we can get back to the world where multiple stops can't be running from different threads things will be much better.  And in that case, you're right that we wouldn't strictly need to worry about making this function re-entrant.  So I'm good with what you have here.

cprice404: I'm expecting the work that you're described to be complete when the current TK PRs and TK-331 are finished.  Do you disagree?

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
camlow325: Ah, k.  I didn't realize Jetty was catching the exception.  I'd been thinking Trapperkeeper was.  Yeah, I agree that testing for log messages might be the best we can do in this case, then.

cprice404: Yeah, I didn't either until I started working on this ticket.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
camlow325: I don't disagree.  That sounds right.

rlinehan:
:+1: once there's a new TK version

cprice404:
I added a commit that simplifies a test based on a new utility macro in the TK logging testutils.

I'm planning on making sure that both this PR and the latest HEAD of TK work ok with puppetserver, and then will do the TK release and update this PR to get rid of the SNAPSHOT.

cprice404:
Rebased, got rid of SNAPSHOT, AFAIK this is ready for merge

camlow325:
:+1:

PR #141, opened 2016-04-04
(TK-124) Add option for session id caching

fpringvaldsen:
Add a new option, `:cache-session-id`, that defaults to true
and allows the caching of the SSL session id.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
@@ -121,7 +121,8 @@
    (schema/optional-key :gzip-enable)                schema/Bool
    (schema/optional-key :access-log-config)          schema/Str
    (schema/optional-key :shutdown-timeout-seconds)   schema/Int
-   (schema/optional-key :post-config-script)         schema/Str})
+   (schema/optional-key :post-config-script)         schema/Str
+   (schema/optional-key :cache-session-id)           schema/Bool})
 
 (def MultiWebserverRawConfigUnvalidated
   {schema/Keyword  WebserverRawConfig})
@@ -178,7 +179,8 @@
    :client-auth                        WebserverSslClientAuth
    (schema/optional-key :ssl-crl-path) (schema/maybe schema/Str)
    :cipher-suites                      [schema/Str]
-   :protocols                          (schema/maybe [schema/Str])})
+   :protocols                          (schema/maybe [schema/Str])
+   :cache-session-id schema/Bool})
 
 (def WebserverSslConnector
   (merge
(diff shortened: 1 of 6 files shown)


src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -90,38 +90,39 @@
    (schema/optional-key :follow-links) schema/Bool})
 
 (def WebserverRawConfig
```
haus: This is a lot of whitespace change. Could this be pared down to just the new key?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -400,7 +402,10 @@
             :cipher-suites           (get-cipher-suites-config config)
             :protocols               (get-ssl-protocols-config config)
             :client-auth             (get-client-auth! config)
-            :ssl-crl-path            (get-ssl-crl-path! config)})))
+            :ssl-crl-path            (get-ssl-crl-path! config)
+            :cache-session-id (if (nil? (:cache-session-id config))
```
haus: I think this could just be `(get config :cache-session-id true)`.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -222,7 +223,8 @@
                            ssl-config)
                         :client-auth :none
                         :cipher-suites (or (:cipher-suites ssl-config) config/acceptable-ciphers)
-                        :protocols     (or (:protocols ssl-config) config/default-protocols)}))
+                        :protocols     (or (:protocols ssl-config) config/default-protocols)
+                        :cache-session-id (or (:cache-session-id ssl-config) true)}))
```
haus: I don't think `or` will work here. If `(:cache-session-id ssl-config)` is nil or false the `or` will evaluate to true.

fpringvaldsen: Ah, you're right. I'm also wondering if we should add a proxy test exercising this piece of functionality.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -222,7 +223,8 @@
                            ssl-config)
                         :client-auth :none
                         :cipher-suites (or (:cipher-suites ssl-config) config/acceptable-ciphers)
-                        :protocols     (or (:protocols ssl-config) config/default-protocols)}))
+                        :protocols     (or (:protocols ssl-config) config/default-protocols)
+                        :cache-session-id (get ssl-config :cache-session-id true)}))
```
fpringvaldsen: Does anyone have any thoughts on what to do about this? Currently a user can configure `:cache-session-id` when adding a proxy route. This option didn't exist before for the proxy route configuration, but we also don't have a way to test it (from what I can tell, there's no way to access the SSL Context Factory for the proxy servlet from the test code, and a more substantial test has proved to be somewhat of a timesink). I'm wondering if we should just hardcode it to true here and use the default behavior?

fpringvaldsen:
@haus I've gone ahead and addressed your changes

haus:
:+1:

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
cprice404: I think I would be fine, personally, with not touching this stuff for the proxy servlet for now, and addressing that in a follow-up ticket if we think it's necessary in the future.  I don't think the proxy stuff would come into play at all for the use cases where CS has run into session issues so far.

PR #142, opened 2016-04-14
Revert "(TK-124) Add option for session id caching"

KevinCorcoran:
Reverts puppetlabs/trapperkeeper-webserver-jetty9#141

@camlow325 Here's this if we need/want it.  I just realized we hadn't done it yet and @haus showed me the cool little button Github has to revert.


haus:
:+1: looks like a revert to me

PR #144, opened 2016-05-04
(TK-372) Reset MBeanContainer's map of bean names.

aperiodic:
In the stop method of the WebserverService, clear the MBeanContainer
class's static map of bean name fragments, which it uses to ensure that
bean names are unique when a name is not specified as the bean is
registered. Without this change, this static map leaks these name
fragments unless JMX metrics are specifically disabled in the TK config.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_service.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_service.clj
@@ -1,4 +1,5 @@
 (ns puppetlabs.trapperkeeper.services.webserver.jetty9-service
+  (:import (org.eclipse.jetty.jmx MBeanContainer))
   (:require
     [clojure.tools.logging :as log]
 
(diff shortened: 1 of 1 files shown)


cprice404:
+1

cprice404:
@rlinehan this is related to the previous round of metrics leakage that you and I worked on, so perhaps you could be the second reviewer?

rlinehan:
@aperiodic this looks fine to me; is there any way we could add a test of this? or is there no good way to write such a test?

aperiodic:
@rlinehan writing a test for this is not easy. The map we're trying to ensure is cleared is private, so it can't be accessed by anything outside the class. I don't know of any remotely straightforward way to get ahold of it to ensure it's cleared. Even if we decided we just wanted to check whether the method is called, since it's a static method of the class, I'm not sure how we'd go about doing that either.

aperiodic:
@rlinehan to guard against this regressing in the future, I added a comment explaining why that method is being called.

cprice404:
@aperiodic if you want a release soon so that you can get this into your app, let's sync up in hipchat today and we can get that going.

rlinehan:
retroactive +1, thanks for adding the comment @aperiodic, and for the explanation.

PR #146, opened 2016-05-09
(maint) bump java-jmx dependency to 0.3.1

wkalt:
This eliminates a conflict with the status and metrics services.


cprice404:
+1

PR #156, opened 2016-12-15
Add uri-path to websocket adapter

wkalt:
Adds uri-path to the WebsocketProtocol and implementation. This caches
the URI used in the upgrade request to establish a websocket connection
so we can retrieve it even after the session has closed. It also caches
the remote address for the same purpose.

### DIFF
--- a/src/puppetlabs/trapperkeeper/services/webserver/experimental/jetty9_websockets.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/experimental/jetty9_websockets.clj
@@ -50,13 +50,16 @@
     (.. this (getSession) (getUpgradeRequest) (isSecure)))
   (peer-certs [this]
     (.. this (getCerts)))
+  (request-path [this]
+    (.. this (getRequestPath)))
   (idle-timeout! [this ms]
     (.. this (getSession) (setIdleTimeout ^long ms)))
   (connected? [this]
     (. this (isConnected))))
 
 (definterface CertGetter
-  (^Object getCerts []))
+  (^Object getCerts [])
+  (^String getRequestPath []))
 
 (defn no-handler
   [event & args]
@@ -65,7 +68,8 @@
 
 (schema/defn ^:always-validate proxy-ws-adapter :- WebSocketAdapter
   [handlers :- WebsocketHandlers
-   x509certs :- [X509Certificate]]
+   x509certs :- [X509Certificate]
+   requestPath :- String]
   (let [{:keys [on-connect on-error on-text on-close on-bytes]
          :or {on-connect (partial no-handler :on-connect)
               on-error   (partial no-handler :on-error)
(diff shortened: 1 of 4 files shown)


src/puppetlabs/trapperkeeper/services/webserver/experimental/jetty9_websockets.clj
```diff
@@ -50,13 +51,16 @@
...
+    (.. this (getURI) (getPath)))
   (idle-timeout! [this ms]
     (.. this (getSession) (setIdleTimeout ^long ms)))
   (connected? [this]
     (. this (isConnected))))
 
 (definterface CertGetter
-  (^Object getCerts []))
+  (^Object getCerts [])
+  (^Object getURI []))
```
camlow325: Would it be useful to type this as `^URI`?

MikaelSmith: Seems like it wouldn't hurt.

MikaelSmith:
The commit message doesn't match the code, it's not saving headers.

src/puppetlabs/trapperkeeper/services/webserver/experimental/jetty9_websockets.clj
```diff
@@ -50,13 +51,16 @@
     (.. this (getSession) (getUpgradeRequest) (isSecure)))
```
MikaelSmith: While we're at it, it'd be nice to stash the remote address as well. We can't log it in on-close because the session goes away.

MikaelSmith:
There aren't really tests for the methods. I'll look at adding some.

MikaelSmith:
I've updated this and added some tests.

MikaelSmith:
Let me just confirm it still seems to work in pcp-broker.

MikaelSmith:
Ok, I'm happy with this iteration.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
```diff
@@ -191,11 +191,18 @@
             server-binary-messages (atom [])
             client-messages        (atom [])
             client-binary-messages (atom [])
+            client-request-path    (atom "")
+            client-remote-addr     (atom "")
+            client-is-ssl          (atom false)
```
camlow325: Maybe it would be good to have the default be something that the `ssl?` function wouldn't return, like `nil`?  Then if the `on-connect` handler did not get invoked, the assertion comparing `client-is-ssl` to `false` would blow up?  Suppose some of the other assertions would blow up anyway in that case, so probably not a big deal.

src/puppetlabs/trapperkeeper/services/webserver/experimental/jetty9_websockets.clj
MikaelSmith: I had something here before, but decided to remove it to avoid changing behavior. It wasn't really needed.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
MikaelSmith: Ah, good idea.

nicklewis:
This makes sense to me.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_handlers_test.clj
MikaelSmith: Pushed an update.

MikaelSmith:
Oh, this should probably have an update to README. I'll put up another PR.

PR #158, opened 2016-12-16
Prepare for 1.6.0 release

MikaelSmith:


camlow325:
@MikaelSmith Ah, hm, noticed we had a Travis failure on your PR:

```
lein test :only puppetlabs.trapperkeeper.services.webserver.jetty9-service-handlers-test/websocket-test

FAIL in (websocket-test) (jetty9_service_handlers_test.clj:256)
can close without supplying a reason
expected: (= 1000 (clojure.core/deref closed))
  actual: (not (= 1000 1006))
```

Probably a transient though and not related to the other changes from your websocket PR, right - #156?

MikaelSmith:
I would expect transient. I don't see anything I changed that would affect that, unless it's relying on timing to get the right response.

camlow325:
Yeah, not sure.  Seems like the test is encountering an abnormal socket close but, yeah, doesn't seem related to your other PR.  So I'm going to go ahead and merge this.  I'll file a ticket about that test failure if I see it come up again.

PR #160, opened 2017-01-10
(TK-429) Allow gzip encoding when an access log is configured

camlow325:
Previously, when the `access-log-config` setting was specified in a
webserver config, gzip encoding was always disabled for responses (even
when `gzip-enabled` was set to true via configuration).  With the
changes in this PR, requests are routed through the GzipHandler whether
or not an access log is configured, allowing the `gzip-enabled` setting
to be properly honored regardless of the value of the
`access-log-config` setting.


src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -626,12 +626,11 @@
     (.setHandlers hc (into-array Handler [(:handlers webserver-context)]))
     (let [shutdown-timeout (when (not (nil? (:shutdown-timeout-seconds options)))
                              (* 1000 (:shutdown-timeout-seconds options)))
-          maybe-zipped (if (or (not (contains? options :gzip-enable))
-                               (:gzip-enable options))
+          maybe-zipped (if (:gzip-enable options true)
```
camlow325: This line wasn't strictly necessary for the fix - just seems cleaner.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -626,12 +626,11 @@
...
                          (gzip-handler hc)
                          hc)
           maybe-logged (if log-handler
-                         (doto log-handler (.setHandler hc))
+                         (doto log-handler (.setHandler maybe-zipped))
```
camlow325: This was the real problem.  We weren't passing in the handler layered with the `GzipHandler` when the log-handler was configured.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -642,20 +642,23 @@
...
+                    :access-log-config
+                    "./dev-resources/puppetlabs/trapperkeeper/services/webserver/request-logging.xml"}}
+       (http-get "http://localhost:8080/hi_world/")
+       ; Logging is done in a separate thread from Jetty and this test. As a result,
+       ; we have to sleep the thread to avoid a race condition.
+       (Thread/sleep 10)
+       (let [list (TestListAppender/list)]
+         (is (re-find #"\"GET /hi_world/ HTTP/1.1\" 200 8\n" (first list))))))
+    (finally
+      (.clear (TestListAppender/list)))))
```
camlow325: The `TestListAppender` uses a static `List` to store logged request info.  This line clears out the list before the test ends so that other tests which use the handler don't end up seeing leftover messages logged from other tests.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
haus: 🐳 did not realize `(:keyword map)` had the default return as the third arg. TIL.

camlow325: I could put a `get` in front of it if you think that'd look better but, yeah, it "just works" with this.

haus: i don't care much. i just didn't know they had the same behavior.

PR #161, opened 2017-01-10
(SERVER-1695) Expose a request-body-max-size webserver setting

camlow325:
This commit exposes a new webserver setting, request-body-max-size,
which can be used to configure the maximum HTTP Content-Length that
the Jetty webserver will accept for an incoming request.  If the
limit is exceeded, the request fails with a 413 error.  If the
setting is not present in configuration or the Content-Length
header is not present in the request, the request will be passed through
to underlying handlers for evaluation.

### DIFF
--- a/doc/jetty-config.md
+++ b/doc/jetty-config.md
@@ -70,6 +70,15 @@ queue is full would be seen by the client as having initially connected to the
 server socket at the TCP layer but having been closed almost immediately
 afterward by the server with no HTTP layer response body.
 
+### `request-body-max-size`
+
+This sets the maximum size, in bytes, of the body for an HTTP request. The size
+of the request body is determined from the value for the request's HTTP
+Content-Length header. If the Content-Length exceeds the configured value, Jetty
+will return an HTTP 413 Error response. If this setting is not configured and/or
+the request does not provide a Content-Length header, Jetty will pass the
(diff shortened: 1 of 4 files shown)


camlow325:
@Sharpie fyi

doc/jetty-config.md
```diff
@@ -70,6 +70,15 @@ queue is full would be seen by the client as having initially connected to the
...
 
+### `request-body-max-size`
+
+This sets the maximum size of the body for an HTTP request. The size of the
+request body is determined from the value for the request's HTTP Content-Length
+header. If the Content-Length exceeds the configured value, Jetty will
+return an HTTP 413 Error response. If this setting is not configured and/or
+the request does not provide a Content-Length header, Jetty will pass the
+request through to underlying handlers (bypassing Content-Length evaluation).
+
```
jpinsonault: Should this mention the units the setting is in?

jpinsonault:
👍 This looks good to me, just one question

doc/jetty-config.md
camlow325: Yeah, good suggestion.  In cd2be9e, I amended the previous commit with this:

> This sets the maximum size, *in bytes*, of the body....

jpinsonault:
I'm 👍

PR #166, opened 2017-02-28
(TK-369) Update jetty dependency to 9.4.1

haus:
This commit updates jetty to 9.4.1, which requires several changes to
tests and the config and core namespaces. Here are a list of changes and
why they were made:

Code changes

* Manually start the request logger when set
  This change
    eclipse/jetty.project@34a8da2
  seems to have caused logging to no longer work without starting here.
  Starting after the server was started did not do the right thing.

* GzipHandler changes
  GzipHandler moved into a new package, into jetty-server from
  jetty-servlets and some of the setup methods changed, notably
  around excluding mime types.

* ProxyServlet changes
  Several methods in the ProxyServlet and AbstractProxyServlet were
  deprecated, so the methods being proxied also have been updated.

* Enable client redirects
  The http connection used for proxies began clearing all handlers in
  eclipse/jetty.project@c7cff6e#diff-571b48f4ff5fbf0a57c41ad7c7ac5dedR297
  so we must manually add the redirect handler back on to correctly
  handle 302s

* Disable symlinks
  In eclipse/jetty.project@d8e6331,
  symlink aliases were allowed by default on all unix platforms, so we
  clear aliases when follow-symlinks is false.

Test changes

* 431 now returned for headers that are too long instead of 413

* SSLException thrown instead of ConnectException for some errors

* URIUtil decodePath changes
  decodePath was changed during a refactor. Path traversal is still
  protected against, but the characters returned have changed, and
  additionally semicolons no longer terminate parsing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,18 @@
 ## 2.0.0
 
-The is a major release
+This is a major release
 
 * [TK-369](https://tickets.puppetlabs.com/browse/TK-369) Move Jetty dependency to 9.4.1
+  Notably, this change drops Java 7 support as Jetty 9.3 and higher only
+  support Java 8 and higher. No public APIs have changed in this release.
+
+* The behavior of normalize-uri-path has changed slightly due to changes in
+  Jetty's URIUtil decodePath method. Specifically, overlong encodings change
+  their decoded value (but are still not traversable), and semicolons no longer
+  terminate URI processing, but only when ; is followed at some point by
+  another path segment beginning with /. For example, "/foo/bar;bar=chocolate/baz;baz=bim"
+  now decodes to "/foo/bar/baz" while before it decoded to "/foo/bar".
+  (https://github.com/eclipse/jetty.project/commit/7f62f2600b943b9aed0e4771891939bf61372c5a)
 
 ## 1.7.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+## 2.0.0
+
+The is a major release
+
+* [TK-369](https://tickets.puppetlabs.com/browse/TK-369) Move Jetty dependency to 9.4.1
+
 ## 1.7.0
 
 This is a feature and bugfix release.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,14 @@ This is a major release
   support Java 8 and higher. No public APIs have changed in this release.
 
 * The behavior of normalize-uri-path has changed slightly due to changes in
-  Jetty's URIUtil decodePath method. Specifically, overlong encodings change
-  their decoded value (but are still not traversable), and semicolons no longer
-  terminate URI processing, but only when ; is followed at some point by
-  another path segment beginning with /. For example, "/foo/bar;bar=chocolate/baz;baz=bim"
-  now decodes to "/foo/bar/baz" while before it decoded to "/foo/bar".
+  Jetty's URIUtil decodePath method. Specifically, overlong encodings (which
+  are considered invalid UTF-8) change their decoded value (but are still not
+  traversable), and semicolons no longer terminate URI processing, but only
+  when ; is followed at some point by another path segment beginning with /.
+  For example, "/foo/bar;bar=chocolate/baz;baz=bim" now decodes to
+  "/foo/bar/baz" while before it decoded to "/foo/bar". If the request has
+  invalid % encoded UTF-8 characters, the path will be decoded as an ISO-8859-1
+  encoded string.
   (https://github.com/eclipse/jetty.project/commit/7f62f2600b943b9aed0e4771891939bf61372c5a)
 
 ## 1.7.0
--- a/doc/jetty-config.md
+++ b/doc/jetty-config.md
@@ -24,10 +24,13 @@ virtual cores on the host divided by 8, with a minimum of 1 and maximum of 4.
 
 ### `selector-threads`
 
-This sets the number of threads that the webserver will dedicate to processing
-events on connected sockets for _unencrypted_ HTTP traffic.  Defaults to the
-number of virtual cores on the host divided by 2, with a minimum of 1 and
-maximum of 4.
+This sets the number of selectors that the webserver will dedicate to
+processing events on connected sockets for unencrypted HTTPS traffic. Defaults
+to the number of virtual cores on the host divided by 2, with a minimum of 1
+and maximum of 4. The number of selector threads actually used by Jetty is
+twice the number of selectors requested. For example, if a value of 3 is
+specified for the `selector-threads` setting, Jetty will actually use 6
+selector threads.
 
 ### `max-threads`
 
@@ -126,10 +129,12 @@ virtual cores on the host divided by 8, with a minimum of 1 and maximum of 4.
 
 ### `ssl-selector-threads`
 
-This sets the number of threads that the webserver will dedicate to processing
-events on connected sockets for _encrypted_ HTTPS traffic.  Defaults to the
-number of virtual cores on the host divided by 2, with a minimum of 1 and
-maximum of 4. The number of selectors used by Jetty is twice the number of ssl
+This sets the number of selectors that the webserver will dedicate to
+processing events on connected sockets for encrypted HTTPS traffic. Defaults
+to the number of virtual cores on the host divided by 2, with a minimum of 1
+and maximum of 4. The number of selector threads actually used by Jetty is
+twice the number of selectors requested. For example, if a value of 3 is
+specified for the `ssl-selector-threads` setting, Jetty will actually use 6
 selector threads.
 
 ### `ssl-cert`
--- a/doc/jetty-config.md
+++ b/doc/jetty-config.md
@@ -82,7 +82,7 @@ request through to underlying handlers (bypassing Content-Length evaluation).
 ### `request-header-max-size`
 
 This sets the maximum size of an HTTP Request Header. If a header is sent
-that exceeds this value, Jetty will return an HTTP 413 Error response. This
+that exceeds this value, Jetty will return an HTTP 431 Error response. This
 defaults to 8192 bytes, and only needs to be configured if an exceedingly large
 header is being sent in an HTTP Request.
 
@@ -129,7 +129,8 @@ virtual cores on the host divided by 8, with a minimum of 1 and maximum of 4.
 This sets the number of threads that the webserver will dedicate to processing
 events on connected sockets for _encrypted_ HTTPS traffic.  Defaults to the
 number of virtual cores on the host divided by 2, with a minimum of 1 and
-maximum of 4.
+maximum of 4. The number of selectors used by Jetty is twice the number of ssl
+selector threads.
 
 ### `ssl-cert`
 
--- a/java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/LifeCycleImplementingRequestLogImpl.java
+++ b/java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/LifeCycleImplementingRequestLogImpl.java
@@ -1,0 +1,32 @@
+package com.puppetlabs.trapperkeeper.services.webserver.jetty9.utils;
+
+import ch.qos.logback.access.jetty.RequestLogImpl;
+import org.eclipse.jetty.util.component.LifeCycle;
+
+/*
+	Sit down, it's story time.
+	Once upon a time logback had a RequestLogImpl that you could just drop in
+	Jetty and use without any modifications. Developers used this and it was
+	good. Then, during Jetty 9.3 development, the RequestLog interface that
+	RequestLogImpl implemented was "refactored"[0] to no longer extend Jetty's
+	LifeCycle interface (this is distinct from Logback's LifeCycle interface,
+	so try to keep up).
+	Implementing Jetty's LifeCycle interface turns out to be important because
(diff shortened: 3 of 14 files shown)


project.clj
```diff
@@ -1,4 +1,4 @@
-(def jetty-version "9.2.10.v20150310")
+(def jetty-version "9.4.1.v20170120")
 
 (defproject puppetlabs/trapperkeeper-webserver-jetty9 "1.7.1-SNAPSHOT"
```
haus: oops.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.0.0
+
+The is a major release
+
+* [TK-369](https://tickets.puppetlabs.com/browse/TK-369) Move Jetty dependency to 9.4.1
```
haus: I don't think there are any user facing changes, so I'm not sure what else to put into the changelog here.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 2.0.0
+
+The is a major release
```
KevinCorcoran: s/the/this

haus: indeed

CHANGELOG.md
KevinCorcoran: How can a major/X release have nothing user facing?  Probably worth at least briefly mentioning the breaking changes to tk-jetty9's API.

haus: I *think* jetty considers them to not be major releases, but it's hard to get at from their docs. I guess the java 7 drop is probably the biggest user facing change.

KevinCorcoran: We're talking about different things.  I was talking about the version of *this* project which you bumped to 2.0.0.

haus: ahhhhh. very well then. i think as of now that is java 7 support ending. unless more changes land before we actually cut 2.0.0. but i can mention that no APIs have changed.

src/puppetlabs/trapperkeeper/services/webserver/normalized_uri_helpers.clj
```diff
@@ -59,14 +59,8 @@
...
-      (let [handler (proxy-super getHandler)
-            is-started (proxy-super isStarted)]
-        ;; It may not strictly be necessary to check if the wrapping handler
-        ;; is started in order to let a request through but that's what the
-        ;; base `HandlerWrapper` class from Jetty does, so it seemed best to
-        ;; follow the pattern:
-        ;; https://github.com/eclipse/jetty.project/blob/jetty-9.2.10.v20150310/jetty-server/src/main/java/org/eclipse/jetty/server/handler/HandlerWrapper.java#L95
-        (when (and handler is-started)
+      (let [handler (proxy-super getHandler)]
+        (when handler
```
KevinCorcoran: could use `when-let`

CHANGELOG.md
```diff
@@ -1,3 +1,19 @@
...
+  Notably, this change drops Java 7 support as Jetty 9.3 and higher only
+  support Java 8 and higher. No public APIs have changed in this release.
+
+* The behavior of normalize-uri-path has changed slightly due to changes in
+  Jetty's URIUtil decodePath method. Specifically, overlong encodings change
+  their decoded value (but are still not traversable), and semicolons no longer
+  terminate URI processing, but only when ; is followed at some point by
+  another path segment beginning with /. For example, "/foo/bar;bar=chocolate/baz;baz=bim"
+  now decodes to "/foo/bar/baz" while before it decoded to "/foo/bar".
+  (https://github.com/eclipse/jetty.project/commit/7f62f2600b943b9aed0e4771891939bf61372c5a)
```
camlow325: Huh, weird.  Oh, well.  Probably good to add/update the blurb about this to the [request-uri-normalization](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9#request-uri-normalization) section in the README.md doc.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -702,8 +704,9 @@
          enable-trailing-slash-redirect? (:enable-trailing-slash-redirect? options)
          normalize-request-uri? (:normalize-request-uri? options)]
      (.setBaseResource handler (Resource/newResource base-path))
-     (when follow-links?
-       (.addAliasCheck handler (AllowSymLinkAliasChecker.)))
+     (if follow-links?
+       (.addAliasCheck handler (AllowSymLinkAliasChecker.))
```
camlow325: Would it be better here to call `.setAliasChecks` with a list containing just `AllowSymLinkAliasChecker`?  Wasn't sure what would happen if you end up adding more than one instance of the same alias checker to the list - like for the case where there's an `AllowSymLinkAliasChecker` that's already been added to the handler by default.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_default_config_test.clj
```diff
@@ -80,36 +80,39 @@ react accordingly."
...
 
 (def selector-thread-count
   "The number of selector threads that should be allocated per connector.  See:
-   https://github.com/eclipse/jetty.project/blob/jetty-9.2.10.v20150310/jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java#L229"
-  (max 1 (min 4 (int (/ (ks/num-cpus) 2)))))
+   https://github.com/eclipse/jetty.project/blob/jetty-9.4.1.v20170120/jetty-server/src/main/java/org/eclipse/jetty/server/ServerConnector.java#L223
+   and
+   https://github.com/eclipse/jetty.project/blob/jetty-9.4.1.v20170120/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java#L403-L408
+   The number of selectors is twice the number of selector threads."
+  (* 2 (max 1 (min 4 (int (/ (ks/num-cpus) 2))))))
```
camlow325: Would be good to update this in the description of the [ssl-selector-threads](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/1.7.0/doc/jetty-config.md#ssl-selector-threads) webserver option too.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -529,7 +532,8 @@
       (tk-app/stop app))))
 
 (deftest large-request-test
-  (testing (str "request to Jetty fails with a 413 error if the request header "
+  ;; This changed from 413 to 431 in https://github.com/eclipse/jetty.project/commit/e53ea55f480a959a2f1f5e2dbdbfc689d61c94a6
+  (testing (str "request to Jetty fails with a 431 error if the request header "
```
camlow325: Would be good to update the expected HTTP status code change for this in the [request-header-max-size](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/1.7.0/doc/jetty-config.md#request-header-max-size) docs.

test/clj/puppetlabs/trapperkeeper/services/webserver/normalized_uri_helpers_test.clj
```diff
@@ -116,8 +116,8 @@
     ;;> or the surrogate pair ED A1 8C ED BE B4 into U+233B4.  Decoding
     ;;> invalid sequences may have security consequences or cause other
     ;;> problems.
-    (is (= "��" (normalize-uri-path-for-string "%C0%AE")))
-    (is (= "/foo/��/��" (normalize-uri-path-for-string "/foo/%C0%AE/%C0%AE")))))
+    (is (= "-64-82" (normalize-uri-path-for-string "%C0%AE")))
```
camlow325: Hmm, this seems unfortunate.  Do you know why this behavior changed?  Ideally, it would not have.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
haus: The follow-links test passes, so jetty seems to handle more than one alias checker just fine, but I'm happy to change it to setAliasCheck instead.

test/clj/puppetlabs/trapperkeeper/services/webserver/normalized_uri_helpers_test.clj
haus: decodePath began catching NotUtf8Exceptions and handling them by falling back to ISO88591 (https://github.com/eclipse/jetty.project/commit/7f62f2600b943b9aed0e4771891939bf61372c5a#diff-3f826383561a9af15d610a1aa21187d9R336).

camlow325: ok, seems a little odd, but I guess that makes some sense.  Would be good to update the comments in this tests and the `normalize-uri-path` docs in the README.md to cover what happens in this case.  Ultimately doesn't seem like a security vulnerability anyway for us since we always put the middleware which uses this normalization in front of our tk-auth middleware so doesn't seem too bad.

haus: It would probably be preferable to raise an exception, but I'm not sure we can easily do that.

haus:
Oh look at the greeeeeeen.

haus:
@camlow325 I think I've responded to all of the feedback so far.

haus:
that travis failure is weird. something about setting stop timeout to 0 and graceful shutdowns, but i couldn't reproduce it locally.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -712,7 +716,7 @@
              in-request-handler (promise)
              ring-handler (fn [_]
                             (deliver in-request-handler true)
-                            (Thread/sleep 300)
+                            (Thread/sleep 2000)
```
camlow325: I think we could bump this up to something much higher, like 30 seconds, to be safer (when running in a super-slow Travis) and still get what we want out of this test.  For the cases that the shutdown happens right away (isn't graceful), this sleep should be interrupted anyway - so a much longer value shouldn't make the test take longer to run.  We could do something here that would be indefinite - like waiting for a promise that will never be delivered.  That might not be good though because in the event the test fails, we'd most likely get a hang.

CHANGELOG.md
camlow325: Could also say here that in the case a non-percent encoded UTF-8 character is found in the path that the path will be percent-decoded using ISO-8859-1 now?

doc/jetty-config.md
```diff
@@ -129,7 +129,8 @@ virtual cores on the host divided by 8, with a minimum of 1 and maximum of 4.
 This sets the number of threads that the webserver will dedicate to processing
 events on connected sockets for _encrypted_ HTTPS traffic.  Defaults to the
 number of virtual cores on the host divided by 2, with a minimum of 1 and
-maximum of 4.
+maximum of 4. The number of selectors used by Jetty is twice the number of ssl
```
camlow325: Looking at [this code segment](https://github.com/eclipse/jetty.project/blob/jetty-9.4.1.v20170120/jetty-server/src/main/java/org/eclipse/jetty/server/Server.java#L403-L408) that you linked, I'm interpreting this to mean that Jetty will allocate 2 threads for each `selector` we request it to use.  We are passing in the value in the config for `selector-threads` and `ssl-selector-threads` as the `selector` argument.  In retrospect, then, I wish we'd just named these config arguments `selectors` and `ssl-selectors`.  I guess that rather than changing the config option name, though, we should say something like this:

 > This sets the number of selectors that the webserver will dedicate to processing events on connected sockets for _encrypted_ HTTPS traffic.  Defaults to the number of virtual cores on the host divided by 2, with a minimum of 1 and maximum of 4.  The number of selector threads actually used by Jetty is twice the number of selectors requested.  For example, if a value of 3 is specified for the `ssl-selector-threads` setting, Jetty will actually use 6 selector threads.

camlow325: Whatever language we end up with for this, it would be good to do the same for the `selector-threads` option - which is the same as this option but for the plaintext port.

haus: seems reasonable.

java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/LifeCycleImplementingRequestLogImpl.java
```diff
@@ -0,0 +1,32 @@
...
+	In order for our RequestLogImpl to automatically start, we decide to extend
+	the existing built-in logback implementation and have it implement Jetty's
+	LifeCycle interface, which it already does, but doesn't declare.
+
+	[0] - https://github.com/eclipse/jetty.project/commit/e3bda4ef
+	[1] - https://github.com/eclipse/jetty.project/blob/0c8273f2ca1f9bf2064cd9c4c939d2546443f759/jetty-util/src/main/java/org/eclipse/jetty/util/component/ContainerLifeCycle.java#L98
+	[2] - https://jira.qos.ch/browse/LOGBACK-1052
+	[3] - https://github.com/qos-ch/logback/pull/269
+
+	And Jetty and Logback lived happily ever after.
```
camlow325: sniff, that was beautiful.

KevinCorcoran: wow i hate open-source software

PR #168, opened 2017-03-29
(TK-437) Reworked no graceful shutdown test to avoid hang

camlow325:
Previously, if the 'no graceful shutdown' test failed due to the server
not closing the socket when the server was stopped, this test could
hang.  This commit reworks the test steps to hopefully avoid hanging the
test suite in the event of this kind of failure.  There does still
appear to be an underlying bug in Jetty here but it has proven very
difficult to reliably reproduce.

### DIFF
--- a/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
+++ b/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
@@ -753,7 +753,7 @@
         ;; isn't closed down even though the server otherwise appears to have
         ;; been stopped properly - leaving the server-side of the connection in
         ;; an indefinite CLOSE_WAIT state.
-        (let [resp @@response
+        (let [resp (deref (deref response))
               error (:error resp)]
           (is (or
                (instance? ConnectionClosedException error)
--- a/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
+++ b/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
@@ -2,7 +2,7 @@
   (:import (org.apache.http ConnectionClosedException)
            (java.io IOException)
            (java.security.cert CRLException)
-           (java.net BindException)
+           (java.net BindException SocketTimeoutException)
            (java.nio.file Paths Files)
            (java.nio.file.attribute FileAttribute)
            (appender TestListAppender)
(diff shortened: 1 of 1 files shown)


camlow325:
Ping @haus for review if you have time.

haus:
i'll take a look soon (TM)

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -706,28 +706,59 @@
...
+        ;; not be completed, which should cause a ConnectionClosedException to
+        ;; be thrown.
+        ;;
+        ;; Note that per TK-437, we have occasionally seen this assertion fail
+        ;; due to a SocketTimeoutException being thrown.  In this case, there
+        ;; appears to be a timing-related bug in Jetty where the server socket
+        ;; isn't closed down even though the server otherwise appears to have
+        ;; been stopped properly - leaving the server-side of the connection in
+        ;; an indefinite CLOSE_WAIT state.
+        (let [resp @@response
```
haus: Why the double @ ?

haus: Oh. the async client returns a promise which is in an atom now?

haus:
Not at all terrifying.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
camlow325: Yeah, I stuck it in the atom so that I could have it make the request while the app was still running but deref the promise after the app has stopped / gone out of scope.  Mostly, I was just doing this to avoid the need to do the double-stop on the webserver which was happening before - which seemed worse than adding the extra state.  I don't think this part of the change really affects the rate of failure for the test anyway, though, so I could move back to making the explicit stop call like before if you think that part makes more sense.

camlow325:
Well, that's good to hear.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
haus: nope, don't care. i just hadn't seen `@@` before and was trying to wrap my head around it.

rlinehan: eek. uh... maybe to make it clearer it would be better to do `(deref (deref ))` or something? I totally thought `@@` might be some special symbol too.

camlow325: I can do `(deref (deref ...))` if you think that's more understandable.

rlinehan: I think I'd prefer that, yeah.

camlow325: Made that change in 6dcc1fa.

PR #169, opened 2017-03-30
(PE-19979) Update i18n to 0.8.0 and clj-parent to 0.6.1

haus:


haus:
once merged, i'll open a merge-up to master

PR #170, opened 2017-04-13
(TK-373) Enable Diffie-Hellman cipher suites

adrienthebo:
This commit enables cipher suites that use classical and elliptic curve
ephemeral Diffie Hellman for key exchange. Per the Mozilla server side
TLS recommendations ECDHE ciphers are preferred, then EDH, and then we
fall back to the original cipher list.

/cc @puppetlabs/security-review


camlow325:
It would be good to update the section of the [docs](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/bcc7c6609946b72b7f1d7da85b2cafc30a4a0a1c/doc/jetty-config.md#cipher-suites) where we list the defaults.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -52,23 +52,22 @@
...
-;;; to be able to add the DH ciphers back in at some point, but we can't do that
-;;; until we're certain that our minimum supported JDK versions for all of our
-;;; supported distros will contain the relevant fixes.
-;;;
 (def acceptable-ciphers
-  ["TLS_RSA_WITH_AES_256_CBC_SHA256"
+  ["TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA256"
+   "TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA"
+   "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256"
+   "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA"
```
joshcooper: Which curves are used with ECDHE? Is it configurable?

adrienthebo: Negotiating a connection with ECDHE indicates the NIST P-256 curve:

```
$  openssl s_client -connect localhost:8081 -cipher ECDHE
[...]
Server Temp Key: ECDH, P-256, 256 bits
---
SSL handshake has read 2179 bytes and written 339 bytes
---
New, TLSv1/SSLv3, Cipher is ECDHE-RSA-AES128-SHA256
[...]
```

I'll get answers for your other questions shortly.

adrienthebo: First off, my knowledge of the different ECC curves is a bit limited, so take this with a grain of sale.

I haven't been able to track down which curves can specifically be used with ECDHE. There's documentation for ECDSA and ECGOST (https://www.bouncycastle.org/wiki/pages/viewpage.action?pageId=362269) which is a starting point, but it includes a number of curves that (in my uneducated guess) seem to be only used with ECDSA.

The Bouncy Castle ECDHE key exchange implementation (https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/crypto/tls/TlsECDHEKeyExchange.java) references a list of curve types (https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/crypto/tls/TlsECCUtils.java#L35-L39) that look like they'll be used when selecting a curve.

As I was writing this I saw some functions that seemed like they might explain things a bit, I'm going to dive into those deeper before I expand on this comment.

adrienthebo: I think I found it: https://github.com/bcgit/bc-java/blob/master/core/src/main/java/org/bouncycastle/crypto/tls/AbstractTlsClient.java#L128-L131

Assuming that this is the code path used, BouncyCastle only supports secp256r1 and secp384r1, and this is not configurable. I'll continue digging deeper to confirm that this is the case, but off the cuff I think this is the determining factor in selecting curves when using ECDH.

camlow325: I don't believe we have any code in our Trapperkeeper-based apps which configure Bouncycastle as a JSSE provider, so I believe we should be looking at this in terms of what the default SunJSSE provider in the JDK would use.  We're only using Bouncycastle for its PKIX API - helper methods behind the Puppet Server Certificate Authority service in the [jvm-ssl-utils](https://github.com/puppetlabs/jvm-ssl-utils) project.

It appears that [the SupportedEllipticCurvesExtension class](http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/687fd7c7986d/src/share/classes/sun/security/ssl/SupportedEllipticCurvesExtension.java) in the JDK source shows the available curves - P-256 (secp256r1) being preferred - but I'm not finding any references to how these could be user / programmatically configured outside of the JDK.  Is it a problem if we can't find / document a way for users to configure which curves are used?

adrienthebo: SunJDK 8u121 (http://www.oracle.com/technetwork/java/javase/8u121-relnotes-3315208.html) added the `jdk.tls.namedGroups` system property to set available/preferred curves. In any case I was able to negotiate an ECDHE cipher with the NIST P-256 curve which is about as ubiquitous as it gets, so I think this is doable.

camlow325: Ah, good find, @adrienthebo.  @joshcooper - do you think think custom configuration is going to be a common-enough thing users might want to do that we should document that somewhere - or do you think it's sufficient for now to just know that it looks like it would be possible to use this (e.g., in the event that we might want to tighten up the defaults in our default packaging, if needed)?

joshcooper: I'm fine just documenting which curves we do support (via java 8). Knowing that it can be configured via Java system property is something we can keep in mind for the future.

PR #171, opened 2017-04-20
(TK-440) Bump jetty version to 9.4 latest

adrienthebo:


rlinehan:
That graceful shutdown test failed again... I don't remember what the deal with that was, if we had put in a "fix" for it or if we were just rerunning all the time?

```
FAIL in (graceful-shutdown-test) (jetty9_service_test.clj:758)

no graceful shutdown when stop timeout is set to 0

request did not error as expected. response: {:opts {:body nil, :decompress-body true, :as :text, :method :get, :url "http://localhost:8080/hello", :headers {"accept-language" "en-US"}}, :error #error {

 :cause nil

 :via

 [{:type java.net.SocketTimeoutException

   :message nil

   :at [org.apache.http.nio.protocol.HttpAsyncRequestExecutor timeout "HttpAsyncRequestExecutor.java" 375]}]

 :trace
```

camlow325:
@rlinehan ugh, yeah there's not really a "fix" for it.  I reworked the test a bit in #168 to allow the test to fail gracefully without hanging.  There may still be a real bug in Jetty there.  Trying to continue to get at the root cause seemed to have diminishing returns to me, though.

I suppose if we get annoyed enough by this failing too frequently, we could just disable the test and file a follow-up investigation ticket for posterity (noting that we probably wouldn't look into it further unless we saw more real-world manifestations of this)?

adrienthebo:
CI passed on round 2.

rlinehan:
I'm cool with us disabling the ticket if it keeps failing.

I'm +1 on this.

PR #177, opened 2017-07-13
(TK-149) Reload the crl file in the running webservers on change

camlow325:
This commit uses the tk-filesystem-watcher service to register a change
notification for the CRL file given to the Jetty webserver.  On
detection of a change, the Jetty SSLContextFactory is reloaded.  This
allows for the updated CRL file to be used without requiring a full
webserver restart.  Runtime CRL updating now occurs when tk-jetty9 is
used in a webservice stack which includes the optional
tk-filesystem-watcher dependency.

### DIFF
--- a/java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/InternalSslContextFactory.java
+++ b/java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/InternalSslContextFactory.java
@@ -1,0 +1,166 @@
+package com.puppetlabs.trapperkeeper.services.webserver.jetty9.utils;
+
+import org.eclipse.jetty.util.log.Log;
+import org.eclipse.jetty.util.log.Logger;
+import org.eclipse.jetty.util.security.CertificateUtils;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
+
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.SSLServerSocket;
+import javax.net.ssl.SSLSocket;
+import java.io.File;
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.security.cert.CRL;
+import java.util.Collection;
+
+public class InternalSslContextFactory extends SslContextFactory {
+
+    private static int maxTries = 5;
+    private static int sleepInMillisecondsBetweenTries = 1000;
+    private static final Logger LOG =
+            Log.getLogger(InternalSslContextFactory.class);
+
+    private Collection<? extends CRL> _crls;
+
+    @Override
+    protected void checkNotStarted() {
+    }
+
+    @Override
+    protected void doStart() throws Exception {
+        synchronized (this) {
+            load();
+        }
+    }
+
+    @Override
+    protected void doStop() throws Exception {
+        synchronized (this) {
+            unload();
+        }
+        super.doStop();
+    }
+
+    @Override
+    protected Collection<? extends CRL> loadCRL(String crlPath) throws Exception {
+        Collection<? extends CRL> crls;
+
+        synchronized (this) {
+            if (_crls == null) {
+                crls = super.loadCRL(crlPath);
+            } else {
+                crls = _crls;
+            }
+        }
+
+        return crls;
+    }
+
+    private void load() throws Exception {
+        synchronized (this) {
+            super.doStart();
+        }
+    }
+
+    private void unload() throws Exception {
+        synchronized (this) {
+            super.doStop();
+        }
+    }
+
+    @Override
+    public SSLContext getSslContext() {
+        synchronized (this) {
+            return super.getSslContext();
+        }
+    }
+
+    @Override
+    public SSLServerSocket newSslServerSocket(String host,
+                                              int port,
+                                              int backlog)
+            throws IOException {
+        synchronized (this) {
+            return super.newSslServerSocket(host, port, backlog);
+        }
+    }
+
+    @Override
+    public SSLSocket newSslSocket() throws IOException {
+        synchronized (this) {
+            return super.newSslSocket();
+        }
+    }
+
+    @Override
+    public SSLEngine newSSLEngine() {
+        synchronized (this) {
+            return super.newSSLEngine();
+        }
+    }
+
+    @Override
+    public SSLEngine newSSLEngine(String host, int port) {
+        synchronized (this) {
+            return super.newSSLEngine(host, port);
+        }
+    }
+
+    @Override
+    public SSLEngine newSSLEngine(InetSocketAddress address) {
+        synchronized (this) {
+            return super.newSSLEngine(address);
+        }
+    }
+
+    public void reload() throws Exception {
+        synchronized (this) {
+            Exception reloadEx;
+            int tries = maxTries;
+            String crlPath = getCrlPath();
+            File crlPathAsFile = null;
+            long crlLastModified = 0;
+
+            if (crlPath != null) {
+                crlPathAsFile = new File(crlPath);
+                crlLastModified = crlPathAsFile.lastModified();
+            }
+
+            // Try to parse CRLs from the crlPath until it is successful
+            // or a hard-coded number of failed attempts have been made.
+            do {
+                reloadEx = null;
+                try {
+                    _crls = CertificateUtils.loadCRL(crlPath);
+                } catch (Exception e) {
+                    reloadEx = e;
+
+                    // If the CRL file has been updated since the last reload
+                    // attempt, reset the retry counter.
+                    if (crlPathAsFile != null &&
+                            crlLastModified != crlPathAsFile.lastModified()) {
+                        crlLastModified = crlPathAsFile.lastModified();
+                        tries = maxTries;
+                    } else {
+                        tries--;
+                    }
+
+                    if (tries == 0) {
+                        LOG.warn("Failed ssl context reload after " +
+                                maxTries + " tries.  CRL file is: " +
+                                crlPath, reloadEx);
+                    } else {
+                        Thread.sleep(sleepInMillisecondsBetweenTries);
+                    }
+                }
+            } while (reloadEx != null && tries > 0);
+
+            if (reloadEx == null) {
+                unload();
+                load();
+            }
+        }
+    }
+}
--- a/java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/InternalSslContextFactory.java
+++ b/java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/InternalSslContextFactory.java
@@ -17,8 +17,8 @@ import java.util.Collection;
 
 public class InternalSslContextFactory extends SslContextFactory {
 
-    private static int maxTries = 5;
-    private static int sleepInMillisecondsBetweenTries = 1000;
+    private static int maxTries = 25;
+    private static int sleepInMillisecondsBetweenTries = 100;
     private static final Logger LOG =
             Log.getLogger(InternalSslContextFactory.class);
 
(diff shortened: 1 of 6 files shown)


src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -952,6 +956,25 @@
...
                  :default-server (get-default-server-from-config config)))
 
+(schema/defn ^:always-validate reload-crl-on-change!
+  "Reload the CRL file used by the supplied ssl-context-factory whenever any
+  changes to the file occur."
+  [ssl-context-factory :- InternalSslContextFactory
+   watcher :- (schema/protocol watch-protocol/Watcher)]
+  (when-let [crl-path (.getCrlPath ssl-context-factory)]
+    (let [normalized-crl-path (.getCanonicalPath (fs/file crl-path))]
+      (watch-protocol/add-watch-dir! watcher (fs/parent crl-path)
```
camlow325: Having to do a recursive watch of all of the files in the directory that the CRL file resides in is definitely not ideal but this is all which is available in the current trapperkeeper-filesystem-watcher service.  At the very least, it should be possible to enhance the watcher service to only watch the immediate directory rather than recursively watching subdirectories.  Even better would be to be able to just watch the CRL file itself but it's not clear to me that the underlying [Path.register](https://docs.oracle.com/javase/8/docs/api/java/nio/file/Path.html#register-java.nio.file.WatchService-java.nio.file.WatchEvent.Kind...-) method which is used by the service would support watching an individual file - as opposed to a directory.

camlow325: Just to confirm from an experiment that I did with this in Java 8 (on OSX at least), if an attempt is made to register a watcher for a regular file (as opposed to a directory), a `java.nio.file.NotDirectoryException` exception is thrown.  So we probably are stuck with needing to watch the parent directory.  Adding an option to avoid the recursive watch, though, would seem like a good thing for this use case.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_service.clj
```diff
@@ -26,23 +33,38 @@
 (defservice jetty9-service
   "Provides a Jetty 9 web server as a service"
   WebserverService
-  [[:ConfigService get-in-config]]
+  {:required [ConfigService]
+   :optional [FilesystemWatchService]}
```
camlow325: I made this an `optional` dependency rather than a required one in order to not break any applications that don't already bootstrap in the tk-filesystem-watcher service on upgrading to the new Jetty version.  This also provides a meager knob for turning off the filesystem CRL watching if it proves to be problematic for some reason.  Longer term, though, it might make sense to just make this be required / always on - maybe in a future tk-jetty9 major release?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
justinstoller: If tk-fs-watcher doesn't fully meet your needs, would it be valuable to use the Java Watcher API directly?

haus: adding a flag to the add-watch-dir! service function to disable recursing seems like it wouldn't be a bad idea to me?

camlow325: For code reuse, I think it would be reasonable for tk-jetty9 to use the tk-fs-watcher and just enhance tk-fs-watcher with an option for not doing the `recursive` directory watching.  The restriction around being unable to set a watcher at a file rather than a directory level, though, seems like a fundamental issue in the Java watcher API implementation.  We might be able to work around that by layering an API on top of the Java FFI / inotify APIs directly but I think that could be a lot of work.  If we decided it were worthwhile to do that rather than waiting for the Java watcher API to hopefully support it directly, might be better to layer that through the tk-filesystem-watcher API as well for better reusability.

haus: @camlow325 that matches my reading of the java watch API. needs to be a directory unfortunately.

camlow325: For reference, I submitted a ticket for the work to add support for non-recursive directory watching to the tk-filesystem-watcher service, [TK-448](https://tickets.puppetlabs.com/browse/TK-448).  It might be reasonable to consider merging this ahead of that work getting done.

java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/InternalSslContextFactory.java
```diff
@@ -0,0 +1,164 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.security.cert.CRL;
+import java.util.Collection;
+
+public class InternalSslContextFactory extends SslContextFactory {
+
+    private static int maxTries = 5;
+    private static int sleepInMillisecondsBetweenTries = 1000;
```
MosesMendoza: This seems like a long interval - do we expect operations that would prevent us from loading the crl to take a second?

haus: I would doubt it. We could adjust the sleep to something like 100 millis and make maxTries 25 or 50?

MosesMendoza: Sounds good. How about 100 millis and max tries 25? Then also our worst case only takes 2.5 seconds to log a failure instead of 5

haus: seems reasonable

src/puppetlabs/trapperkeeper/services/webserver/jetty9_service.clj
```diff
@@ -26,23 +33,38 @@
...
+               started-context (core/start! context config)]
+           (if-let [filesystem-watcher-service
+                    (maybe-get-service this :FilesystemWatchService)]
+             (let [watcher (watch-protocol/create-watcher filesystem-watcher-service)]
+               (doseq [server (:jetty9-servers started-context)]
+                 (when-let [ssl-context-factory (-> server
+                                                    second
+                                                    :state
+                                                    deref
+                                                    :ssl-context-factory)]
```
MosesMendoza: @haus do you know if/why/how this is using the thread-first macro? I'm only familiar with that macro in the context of passing a "threading" parameter/return values as arguments to success function calls, but this is a single function call. What's actually going on?

justinstoller: It's still passing results to successive function calls, though the macro doesn't require parens unless there's additional parameters (I think), ie it could be re-written like this:
```clj
(let [server-instance (second server)
      state-atom (:state server-instance)
      referenced-state (deref state-atom)]
  (when-let [ssl-context-factory (:ssl-context-factory referenced-state)]
    (...)))
```

haus: I'll try to walk through it. First, `:jetty9-servers` is a map of id, server kv pairs. These pairs can be accessed using vec access calls such as `second`. So the first bit is passing the `server` kv pair to `second`, which returns the server (and drops the id). Inside the server context:

```(def ServerContext
  {:state     (schema/atom ServerContextState)
   :handlers  ContextHandlerCollection
   :server    (schema/maybe Server)})
```

In clojure, using a keyword as a function does a lookup in the map on that keyword, so calling `:state` on the server-context is the equivalent of doing `(get server-context :state)`. Looking at the schema for `:state`, it is an atom holding a `ServerContextState`, so when `deref` is called on the `:state` bit, a ServerContextState is returned. `(deref foo)` is the same as `@foo`.

```(def ServerContextState
  {:mbean-container (schema/maybe MBeanContainer)
   :overrides-read-by-webserver schema/Bool
   :overrides (schema/maybe {schema/Keyword schema/Any})
   :endpoints RegisteredEndpoints
   :ssl-context-factory (schema/maybe InternalSslContextFactory)})
```

The last item in the threading macro is another map lookup, which itself returns an `InternalSslContextFactory`. 
How was that?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_service.clj
```diff
@@ -26,23 +33,38 @@
...
+                   (core/reload-crl-on-change! ssl-context-factory watcher)))
+               (assoc started-context :watcher watcher))
+             started-context)))
```
MosesMendoza: In the previous iteration it looks like we would always return the return value of `(core/start! context config)` which I think was "context". It looks like we've moved the call to `core/start!` inside the `let` definitions. Then within the `if-let` we return the context with the new file system watcher added. But if the `if-let` evaluates to false because we don't have a `FileSystemWatchService` it looks like the function would now return `nil`. Is that right, and if so ok?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_service.clj
MosesMendoza: Awesome, thank you both! Made much more sense when I learned you didn't have to supply the parens.. I hadn't seen it without them so I read this as a single function call instead of the multiple calls I expected. Got it now.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -372,6 +376,55 @@
...
+       (Thread/sleep 1000)
+       (fs/copy "./dev-resources/config/jetty/ssl/crls/crls_localhost_revoked.pem"
+                tmp-file)
+       (loop [times 30]
+         (cond
+           (try
+             (ssl-exception-thrown? (get-request))
+             (catch IllegalStateException _
+               false))
+           (is true)
```
MosesMendoza: Why do we have this line? Seems unnecessary unless i'm misunderstanding something about the way these tests are defined.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_service.clj
haus: The other branch of the `if-let` will return just the `started-context`, which is set to the return of `core/start!` unless I'm mistaken.

MosesMendoza: Oh I think you're right - I had a parens off-by-one and thought the `started-context` was in the true branch of the `if-let`

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
haus: i left a similar comment in the puppetserver PR. It's weird to me to see the assertion inside the loop construct. I would probably move the `is` to be around the `loop`, and then update this to be just `true`. Similarly, the line on 420 would just be `(ssl-exception-thrown? (get-request))`.

haus: If folks feel differently I'd like to hear that though.

@MosesMendoza the `(is true)` is the expr half of the test/expr pair for that cond. the `(try...)` is the test. if it returns something truthy then `(is true)` is evaluated.

camlow325: I can do that.  Not a big deal to me either way.

MosesMendoza: > the (is true) is the expr half of the test/expr pair for that cond. the (try...) is the test. if it returns something truthy then (is true) is evaluated.

Ah I see. I had `cond` wrong in my head. Thanks

haus: @camlow325 I had it on my plate already so I don't mind pushing up a few changes today.

haus:
@MosesMendoza updated

PR #181, opened 2017-07-26
(TK-449) Remove invalid cipher suite

justinstoller:
Previously we included an invalid cipher suite in our list of cipher
suites to register.

This causes a warning to be logged at start up.

This patch removes the invalid cipher suite from the list of cipher
suites we register.

### DIFF
--- a/doc/jetty-config.md
+++ b/doc/jetty-config.md
@@ -234,7 +234,6 @@ you'll need to use the all-caps cipher suite name.
 
 If not supplied, trapperkeeper uses this list of cipher suites:
 
-- `TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA256`
 - `TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA`
 - `TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256`
 - `TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA`
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
@@ -53,8 +53,7 @@
 ;;; risky for our downstream apps, thus it was decided that it makes sense to
 ;;; keep these overrides.
 (def acceptable-ciphers
-  ["TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA256"
-   "TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA"
+  ["TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA"
    "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256"
    "TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA"
 


justinstoller:
I went ahead and created a ticket for it. This is the work mentioned in an email today, @jsane @jpinsonault, @MosesMendoza

adrienthebo:
It looks like there are multiple ciphers (on my JVM instance at least) that might need to be changed or replaced:

```
2017-07-27 12:03:49,421 INFO  [async-dispatch-2] [o.e.j.u.s.SslContextFactory] No Cipher matching 'TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA256' is supported
2017-07-27 12:03:49,421 INFO  [async-dispatch-2] [o.e.j.u.s.SslContextFactory] No Cipher matching 'TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA' is supported
2017-07-27 12:03:49,421 INFO  [async-dispatch-2] [o.e.j.u.s.SslContextFactory] No Cipher matching 'TLS_DHE_RSA_WITH_AES_256_CBC_SHA256' is supported
2017-07-27 12:03:49,421 INFO  [async-dispatch-2] [o.e.j.u.s.SslContextFactory] No Cipher matching 'TLS_DHE_RSA_WITH_AES_256_CBC_SHA' is supported
2017-07-27 12:03:49,421 INFO  [async-dispatch-2] [o.e.j.u.s.SslContextFactory] No Cipher matching 'TLS_RSA_WITH_AES_256_CBC_SHA256' is supported
2017-07-27 12:03:49,421 INFO  [async-dispatch-2] [o.e.j.u.s.SslContextFactory] No Cipher matching 'TLS_RSA_WITH_AES_256_CBC_SHA' is supported
```

None of these are strictly harmful; if the cipher isn't present it won't be used. http://docs.oracle.com/javase/7/docs/technotes/guides/security/StandardNames.html#ciphersuites indicates that some of these cipher names aren't valid, but that AES256 encryption only uses SHA384 or SHA512 ciphers.

justinstoller:
That's wild. My understanding was that we will go down the list until we find a cipher that is implemented in the JVM, that newer versions of the JVM implement more/stronger ciphers, but that the cipher this patch removes was never implemented in any JVM and will log a warning on normal start up everywhere.

I was able to find all of the ciphers listed in your warning on the standard names page you linked, except for the first. Do you know if there's a JVM that implements `TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA256` despite Oracle's documentation?

For curiousity's sake, what JVM and version are you using?

adrienthebo:
```
└> java -version
java version "1.8.0_121"
Java(TM) SE Runtime Environment (build 1.8.0_121-b13)
Java HotSpot(TM) 64-Bit Server VM (build 25.121-b13, mixed mode)
```

I'm pretty sure we could spend a hilarious amount of time picking over cipher suites. Let's definitely merge this PR; I'll recheck the other current ciphers in the mean time.

justinstoller:
Do you have JCE installed?

adrienthebo:
This might be a limitation of the tool that I'm using to install different versions of Java: https://github.com/shyiko/jabba/issues/45

The bikeshed shall be blue.

PR #182, opened 2017-08-04
(TK-448) use non-recursive form of watch service watcher

MosesMendoza:
This commit updates the add-watch-dir! call that we use to watch the crl
directory for changes to use the non-recursive option. We do so to address an
issue we were seeing where during the normal lifecycle of puppetserver run, it
was possible to create a directory that the watch service could not traverse,
causing the service to fail. With recursive false, we do not expect the service
to fail.

Signed-off-by: Moses Mendoza <moses@puppet.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ lein: 2.7.1
 jdk:
   - oraclejdk7
   - openjdk7
+  - oraclejdk8
 script: ./ext/travisci/test.sh
 notifications:
   email: false
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 1.8.1
+
+This is a bugfix release.
+
+* [TK-448](https:/tickets.puppet.com/browse/TK-448) Switch to using
+  non-recursive directory watching for automatic CRL reloading to avoid failure
+  due to unreadable/inaccessible directories in the SSL dir.
+
 ## 1.8.0
 
 This is a feature release.
--- a/project.clj
+++ b/project.clj
@@ -40,7 +40,7 @@
                  [puppetlabs/kitchensink]
                  [puppetlabs/trapperkeeper]
                  [puppetlabs/i18n]
-                 [puppetlabs/trapperkeeper-filesystem-watcher]
+                 [puppetlabs/trapperkeeper-filesystem-watcher "1.1.0"]
                  ]
 
   :source-paths  ["src"]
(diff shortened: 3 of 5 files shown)


MosesMendoza:
Blocked on https://github.com/puppetlabs/trapperkeeper-filesystem-watcher/pull/17

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -400,6 +400,15 @@
...
          (is (= (:body response) hello-body))))
 
+     ;;(testing "request to jetty successful when a non-traversable directory is created in cert dir"
+     ;;   (let [parent (fs/parent tmp-file)
+     ;;         non-traversable-dir (fs/temp-dir "foo")]
+     ;;     (fs/chmod "-rwx" non-traversable-dir)
+     ;;     (fs/sym-link (fs/file parent "bar") non-traversable-dir)
+     ;;     (let [response (get-request)]
+     ;;       (is (= (:status response) 200))
+     ;;       (is (= (:body response) hello-body)))))
```
MosesMendoza: I could only exact a failure of the watch service with :recursive true by adding a breakpoint in this test and creating a directory owned by `root` in crl directory. Just creating a directory with non-traversable permissions is insufficient. Also this symlink business didn't work. I couldn't think of a way to create a directory owned by root in the test that was sane, so I'm wondering if we want to just say the existing validation (which exercises the crl reload functionality) is sufficient.

MosesMendoza:
Test failures are currently expected - this will not work without https://github.com/puppetlabs/trapperkeeper-filesystem-watcher/pull/17 merged first and used by tk-jetty

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
haus: That seems fine. The sort of test that you're describing sounds like it belongs in tk-fs-watcher anyway?

MosesMendoza: Yeah

MosesMendoza: removed

MosesMendoza:
Need to push another change to move the recursive setting to `create-watcher`

PR #183, opened 2017-08-25
(TK-451) Support runtime refresh of CRL

Magisus:
This PR ports the runtime refresh of the CRL from the 1.x branch to master. It includes a no-op mergeup of the 1.x branch.

### DIFF
--- a/java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/InternalSslContextFactory.java
+++ b/java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/InternalSslContextFactory.java
@@ -12,8 +12,8 @@ import java.util.function.Consumer;
 
 public class InternalSslContextFactory extends SslContextFactory {
 
-    private static int maxTries = 5;
-    private static int sleepInMillisecondsBetweenTries = 1000;
+    private static int maxTries = 25;
+    private static int sleepInMillisecondsBetweenTries = 100;
     private static final Logger LOG =
             Log.getLogger(InternalSslContextFactory.class);
     private static Consumer<SslContextFactory> consumer = sslContextFactory
--- a/java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/InternalSslContextFactory.java
+++ b/java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/InternalSslContextFactory.java
@@ -1,0 +1,86 @@
+package com.puppetlabs.trapperkeeper.services.webserver.jetty9.utils;
+
+import org.eclipse.jetty.util.log.Log;
+import org.eclipse.jetty.util.log.Logger;
+import org.eclipse.jetty.util.security.CertificateUtils;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
+
+import java.io.File;
+import java.security.cert.CRL;
+import java.util.Collection;
+import java.util.function.Consumer;
+
+public class InternalSslContextFactory extends SslContextFactory {
+
+    private static int maxTries = 5;
+    private static int sleepInMillisecondsBetweenTries = 1000;
+    private static final Logger LOG =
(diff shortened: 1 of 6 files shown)


Magisus:
I think I know why I was confused about what needed to be ported from Jetty: the version of `InternalSslContextFactory` in the 1.x branch is much more complicated than the one in the Jeremy's PR against master. I believe we do need to take the version in that PR wholesale, as is done here.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -23,7 +23,8 @@
            (java.lang.management ManagementFactory)
            (org.eclipse.jetty.jmx MBeanContainer)
            (org.eclipse.jetty.util URIUtil BlockingArrayQueue)
-           (java.io IOException))
+           (java.io IOException File)
```
justinstoller: Don't believe File is actually used in this namespace

src/puppetlabs/trapperkeeper/services/webserver/jetty9_service.clj
```diff
@@ -1,13 +1,20 @@
 (ns puppetlabs.trapperkeeper.services.webserver.jetty9-service
-  (:import (org.eclipse.jetty.jmx MBeanContainer))
+  (:import (org.eclipse.jetty.jmx MBeanContainer)
+           (java.io File))
```
justinstoller: Don't think File is used in this namespace

project.clj
```diff
@@ -40,6 +40,7 @@
                  [puppetlabs/kitchensink]
                  [puppetlabs/trapperkeeper]
                  [puppetlabs/i18n]
+                 [puppetlabs/trapperkeeper-filesystem-watcher "1.1.0"]
```
justinstoller: Should we release clj-parent first and bump it here instead of fs-watcher?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_service.clj
Magisus: Ack, forgot to remove those when I switched over to cherry-picking these commits.

project.clj
Magisus: Yeah that's probably better. In that case we need to merge https://github.com/puppetlabs/clj-parent/pull/79 first.

Magisus:
Updated with new clj-parent and @justinstoller 's feedback

java/com/puppetlabs/trapperkeeper/services/webserver/jetty9/utils/InternalSslContextFactory.java
```diff
@@ -0,0 +1,84 @@
...
+                crls = super.loadCRL(crlPath);
+            } else {
+                crls = _crls;
+            }
+        }
+
+        return crls;
+    }
+
+    public void reload() throws Exception {
```
MosesMendoza: I thought this was what was already built into Jetty?

Magisus: No, the version in Jetty takes and argument, you can see us calling it at the end of this function.

MosesMendoza: Oh never mind, I see the difference now. That branch had substantially more revision to backport the synchronized state mgmt around the SSL content.

Magisus: Yeah that's the conclusion I arrived at too.

MosesMendoza: yeah and it looks like this version is an improved adaptation to allow us to do more load attempts if something is incorrect with the crl itself

PR #186, opened 2017-09-19
(SERVER-1597) Add support for renegotiationAllowed configuration

jsane:
Introduced a new parameter to disallow renegotiation with default set to true. Done to address security vulnerability that allows a client to DoS any Jetty based puppet service.

This will affect all services using TK-webserver-jetty9. Some limited testing has been done using puppet server. Pl refer to comments in SERVER-1597 and related  tickets if enabling this could adversely impact your service. 
Would like to know if there any acceptance tests that need to be run separately besides this PR.

### DIFF
--- a/doc/jetty-config.md
+++ b/doc/jetty-config.md
@@ -291,6 +291,13 @@ Optional. This describes a path to a Certificate Revocation List file. Incoming
 SSL connections will be rejected if the client certificate matches a
 revocation entry in the file.
 
+### `allow-renegotiation`
+
+Optional. This controls whether the web server will allow client initiated SSL/TLS
+renegotiations. By default this will be disabled since allowing client to renegotiate
+is a vulnerability causing denial of service and information disclosure in certain
+cases. It can be over-ridden by setting this parameter to true.
+
 ### `static-content`
 
 Optional. This is a list of static content to be added to the server as context handlers
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
@@ -68,6 +68,7 @@
    "TLS_RSA_WITH_AES_128_CBC_SHA"])
 (def default-protocols ["TLSv1" "TLSv1.1" "TLSv1.2"])
 (def default-client-auth :need)
+(def default-allow-renegotiation false)
 
 ;;;
 ;;; JMX
@@ -117,7 +118,8 @@
    (schema/optional-key :gzip-enable)                schema/Bool
    (schema/optional-key :access-log-config)          schema/Str
    (schema/optional-key :shutdown-timeout-seconds)   schema/Int
-   (schema/optional-key :post-config-script)         schema/Str})
+   (schema/optional-key :post-config-script)         schema/Str
+   (schema/optional-key :allow-renegotiation)        schema/Bool})
 
 (def MultiWebserverRawConfigUnvalidated
   {schema/Keyword  WebserverRawConfig})
@@ -174,7 +176,8 @@
    :client-auth                        WebserverSslClientAuth
    (schema/optional-key :ssl-crl-path) (schema/maybe schema/Str)
    :cipher-suites                      [schema/Str]
-   :protocols                          (schema/maybe [schema/Str])})
+   :protocols                          (schema/maybe [schema/Str])
+   (schema/optional-key :allow-renegotiation)     (schema/maybe schema/Bool)})
 
 (def WebserverSslConnector
   (merge
@@ -388,7 +391,9 @@
             :cipher-suites           (get-cipher-suites-config config)
             :protocols               (get-ssl-protocols-config config)
             :client-auth             (get-client-auth! config)
-            :ssl-crl-path            (get-ssl-crl-path! config)})))
+            :ssl-crl-path            (get-ssl-crl-path! config)
+            :allow-renegotiation     (get config :allow-renegotiation 
+                                         default-allow-renegotiation)})))
 
 (schema/defn ^:always-validate
   maybe-add-http-connector :- {(schema/optional-key :http) WebserverConnector
--- a/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
+++ b/src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
@@ -82,7 +82,8 @@
 (def ProxySslConfig
   (merge config/WebserverSslPemConfig
          {(schema/optional-key :cipher-suites) [schema/Str]
-          (schema/optional-key :protocols)     (schema/maybe [schema/Str])}))
+          (schema/optional-key :protocols)     (schema/maybe [schema/Str])
+          (schema/optional-key :allow-renegotiations)     (schema/maybe [schema/Bool])}))
 
 (def ProxyOptions
   (assoc CommonOptions
@@ -187,7 +188,7 @@
 (schema/defn ^:always-validate
   ssl-context-factory :- SslContextFactory
   "Creates a new SslContextFactory instance from a map of SSL config options."
-  [{:keys [keystore-config client-auth ssl-crl-path cipher-suites protocols]}
+  [{:keys [keystore-config client-auth ssl-crl-path cipher-suites protocols allow-renegotiation]}
    :- config/WebserverSslContextFactory]
   (if (some #(= "sslv3" %) (map str/lower-case protocols))
     (log/warn (i18n/trs "`ssl-protocols` contains SSLv3, a protocol with known vulnerabilities; we recommend removing it from the `ssl-protocols` list")))
@@ -212,6 +213,9 @@
       :need (.setNeedClientAuth context true)
       :want (.setWantClientAuth context true)
       nil)
+    (if allow-renegotiation
+      (.setRenegotiationAllowed context true)
+      (.setRenegotiationAllowed context false))
     (when ssl-crl-path
       (.setCrlPath context ssl-crl-path)
       ; .setValidatePeerCerts needs to be called with a value of 'true' in
@@ -228,7 +232,9 @@
                            ssl-config)
                         :client-auth :none
                         :cipher-suites (or (:cipher-suites ssl-config) config/acceptable-ciphers)
-                        :protocols     (or (:protocols ssl-config) config/default-protocols)}))
+                        :protocols     (or (:protocols ssl-config) config/default-protocols)
+                        :allow-renegotiation  (or (:allow-renegotiation ssl-config)
+                                                  config/default-allow-renegotiation)}))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;;; Jetty Server / Connector Functions
@@ -357,11 +363,11 @@
         (.addConnector server connector)))
     (when-let [https (:https config)]
       (let [ssl-ctxt-factory (ssl-context-factory
-                               {:keystore-config (:keystore-config https)
-                                :client-auth     (:client-auth https)
-                                :ssl-crl-path    (:ssl-crl-path https)
-                                :cipher-suites   (:cipher-suites https)
-                                :protocols       (:protocols https)})
+                               (select-keys https
+                                            [:keystore-config :client-auth
+                                             :ssl-crl-path :cipher-suites
+                                             :allow-renegotiation
+                                             :protocols]))
             connector        (ssl-connector server ssl-ctxt-factory https)]
 
         (.addConnector server connector)
--- a/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
+++ b/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
@@ -64,6 +64,7 @@
       (update-in [:https :cipher-suites] (fnil identity acceptable-ciphers))
       (update-in [:https :protocols] (fnil identity default-protocols))
       (update-in [:https :client-auth] (fnil identity default-client-auth))
+      (update-in [:https :allow-renegotiation] (fnil identity default-allow-renegotiation))
       (update-in [:https :ssl-crl-path] identity)))
 
 (deftest process-config-http-test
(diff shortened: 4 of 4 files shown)


jsane:
The above test failure appears due to missing lein2 which is not even part of the repo. Is this expected / can be ignored or am I missing something?

jsane:
@Magisus 
Like I had pinged in the server room sometime back, not sure what is causing the travis failure. It fails to find lein2 referred to from within the ext/travisci/test.sh script. And that file has not changed since it was originally checked in (if one were to explain introduction of lein2 in there) and another PR just a few weeks ago passes the travis checks just fine. 
This is not super important but would like to figure/know what may be causing it. It would be nice to be able to enable this option that addresses a security vulnerability. thanks

Magisus:
That's really strange. I rekicked the job, we'll see if it is transient for some reason. If not, maybe try rebasing these commits? Other server PRs seem to be passing these Travis checks, so it doesn't sound like anything is wrong with our Travis config.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -388,7 +391,8 @@
             :cipher-suites           (get-cipher-suites-config config)
             :protocols               (get-ssl-protocols-config config)
             :client-auth             (get-client-auth! config)
-            :ssl-crl-path            (get-ssl-crl-path! config)})))
+            :ssl-crl-path            (get-ssl-crl-path! config)
+            :disallow-renegotiation  (or (:disallow-renegotiation config) default-disallow-renegotiation)})))
```
justinstoller: I assume the end result of a user's actions would cause `(disallow-renegotiation config)` to return `false` if they wanted to change this setting? If they did, wouldn't this still evaluate to `true` because of the default? How does a user enable renegotiation?

justinstoller:
About this improvement generally:
 1. Are there performance implication for disabling renegotiation?
 2. What do we think most folks want "out of the box"?
 3. Is this the current default or is this a change of behavior from current usage? And more importantly, is it backwards compatible with common client configurations?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -117,7 +118,8 @@
    (schema/optional-key :gzip-enable)                schema/Bool
    (schema/optional-key :access-log-config)          schema/Str
    (schema/optional-key :shutdown-timeout-seconds)   schema/Int
-   (schema/optional-key :post-config-script)         schema/Str})
+   (schema/optional-key :post-config-script)         schema/Str
+   (schema/optional-key :disallow-renegotiation)      schema/Bool})
```
justinstoller: Since this was written our style guide has changed to not align values in a map, however to keep with the style of this file I think we should probably do that.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
```diff
@@ -136,19 +136,19 @@
              (merge valid-ssl-pem-config
                     {:ssl-host "foo.local"}))
            (munge-expected-https-config
-             {:https {:host "foo.local" :port default-https-port}})))
+             {:https {:host "foo.local" :port default-https-port :disallow-renegotiation default-disallow-renegotiation}})))
```
justinstoller: This map is now long enough that I think it makes sense to break it up by key-value pairs.

jsane:
1. I am aware of some concerns expressed around possible performance implications but based on my understanding it need not be. It should still be possible to utilize session IDs/tickets, w/o enabling renegotiation, which are some common performance improvement techniques used. Also given that our clients always establish authenticated sessions I am not sure if they require or actively engage in renegotiations. However if there is hint that it might be causing performance issues we can disable it. 
2. Client initiated renegotiations to be disabled. This is a well known security vulnerability and almost universally it is recommended to disallow it. 
3. It would be a change. Ironically or interestingly, jetty9 documentation claims the default for this parameter (renegotiationAllowed) is false. So while it was supposed to be the default behavior in reality it was found to be not the case. Setting it explicitly actually disables it. I would be interested to find if any of our clients have been relying or using it etc. Other than browsers, are there any other clients that we do not control (or know about) like various puppet developed CLIs, agents?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
jsane: Good point and thanks for catching it. Got left out when I switched to using disallow instead of allow semantics midway to avoid using an 'else' in Clojure that I initially struggled with :-)
In any case it is possible to over ride the default of disallowing renegotiations by adding "allow-renegotiation" to true in appropriate configuration files (like pe-puppet-server.conf). 
Do I need to document it someplace say in the main TK-webserver-jetty9 readme ?

justinstoller: > Do I need to document it someplace

Yes. Webserver config docs live [here](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/blob/master/doc/jetty-config.md)

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -388,7 +391,9 @@
             :cipher-suites           (get-cipher-suites-config config)
             :protocols               (get-ssl-protocols-config config)
             :client-auth             (get-client-auth! config)
-            :ssl-crl-path            (get-ssl-crl-path! config)})))
+            :ssl-crl-path            (get-ssl-crl-path! config)
+            :allow-renegotiation     (or (:allow-renegotiation config)
```
justinstoller: If this is set to `false` explicitly in the config and we change the default to `true` users won't be able to override this, right? I think this pattern works above because the config values are expected to *not* be booleans, since we're expecting booleans here we might want to do something more like:

```clj
(get config :allow-renegotiation default-allow-renegotiation)
```

doc/jetty-config.md
```diff
@@ -291,6 +291,13 @@ Optional. This describes a path to a Certificate Revocation List file. Incoming
 SSL connections will be rejected if the client certificate matches a
 revocation entry in the file.
 
+### `allow-renegotiation`
+
+Optional. This controls whether the web server would allow client initiated SSL/TLS
+renegotiations. By default this would be disabled since allowing client to renegotiate
```
justinstoller: I think we want to substitute out "would" with "will"? in these two lines.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -212,6 +212,7 @@
       :need (.setNeedClientAuth context true)
       :want (.setWantClientAuth context true)
       nil)
+    (if allow-renegotiation (.setRenegotiationAllowed context true) (.setRenegotiationAllowed context false))
```
justinstoller: We should break this line for length. I think like?
```
  (if test
    then
    else)
```

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
```diff
@@ -136,19 +136,25 @@
              (merge valid-ssl-pem-config
                     {:ssl-host "foo.local"}))
            (munge-expected-https-config
-             {:https {:host "foo.local" :port default-https-port}})))
+             {:https {:host "foo.local"
+                      :port default-https-port 
+                      :allow-renegotiation default-allow-renegotiation}})))
```
justinstoller: In general having to add this default to all of the https tests seems less than ideal. Does it have to be here for the tests to pass? If so can it and the other defaults be added to a defaults map that is merged with keys that we actually care about in each test?

Also, we should assert that you can set this to a non-default value, I guess. I'd expect to see using default and non-default values in the `munge-actual-https-config` section either an existing, or new, assertion.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
```diff
@@ -168,15 +175,18 @@
            (munge-expected-https-config
              {:https {:host                   "foo.local"
                       :port                   8001
-                      :so-linger-milliseconds 22000}})))
+                      :so-linger-milliseconds 22000
+		      :allow-renegotiation default-allow-renegotiation}})))
```
justinstoller: whitespace

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -361,6 +364,7 @@
                                 :client-auth     (:client-auth https)
                                 :ssl-crl-path    (:ssl-crl-path https)
                                 :cipher-suites   (:cipher-suites https)
+                                :allow-renegotiation (:allow-renegotiation https)
```
justinstoller: Nice to have: using [`select-keys`](https://clojuredocs.org/clojure.core/select-keys) instead of manual map construction.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -228,7 +229,9 @@
                            ssl-config)
                         :client-auth :none
                         :cipher-suites (or (:cipher-suites ssl-config) config/acceptable-ciphers)
-                        :protocols     (or (:protocols ssl-config) config/default-protocols)}))
+                        :protocols     (or (:protocols ssl-config) config/default-protocols)
+                        :allow-renegotiation  (or (:allow-renegotiation ssl-config)
```
justinstoller: This has the same issue as [above](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/186/files#r144926581). I'm curious as to why we're setting this in the config and here? If you know off the top of your head I'd love to hear it, otherwise I'll look deeper after lunch.

justinstoller:
@jsane you're reply makes sense to me 👍, though there's still an issue with the code. Give it a once over and I'm happy to dive into some of it with if you'd like.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
jsane: My Clojure skills are rudimentary at best so my changes are not the most elegant but would be happy to adjust once I understand/figure out the above construct.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
jsane: About the ability to over ride the default (disabling renegotiations). I think the latest update should allow it to enable renegotiations by adjusting it in config file. At least I was able to in my testing and it seems to work as intended but I too am curious to know what case could be missed.  
About setting it in config and above, I just mimicked what is being done for other configuration settings like protocols, cipher-suites etc.   But if you think there is some redundancy I can experiment and figure the minimal change set to achieve the desired functionality.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
jsane: Just noticed this comment.. let me ponder upon it and get back.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
justinstoller: Here we're just building a smaller map to pass into `ssl-context-factory`, using as far as I can tell, a subset of the keys (without changing any names) from the `https` map. Select-keys will do exactly that for us like:
```clj
  (let [ssl-ctxt-factory (ssl-context-factory
                           (select-keys https
                                       [:keystore-config :client-auth
                                        :ssl-crl-path :cipher-suites
                                        :allow-renegotiation]))])
```

```

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
justinstoller: I dove into this a bit and I think that we should update [ProxSslConfig](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/186/files/8c457c5941862d6af036dafa73f374903497853b#diff-1fc7ef9acd5fbc52947b0a4ba167e727R82) above to include an optional key for `:allow-renegotiations`. I also think we should look at making that key required in [`config/WebserverSslContextFactory`](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/186/files/8c457c5941862d6af036dafa73f374903497853b#diff-83ec7cd15abf3ccb0d9023ae974afc20R174).

I can go into why I think that, but I spent the last while writing a tome that I think might actually just obscure the actionable part.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
jsane: I agree your proposal promotes genericity though I would not think (or have liked) allowing renegotiation by default as possible usage except in some completely security agnostic situations etc. In any case I have done the needful.   thanks

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_config_test.clj
jsane: I have made some updates. pl let me know if I missed any test cases.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
jsane: I have left the new key optional in config/WebserverSslContextFactory. One reason being it would require specifying it in all config files of the different projects using TK-webserver-jetty9 and just fear that someone might set it to true. Or else I would need to change all the default config files for the repo's using TK. 
Instead given the motivation for introducing this configuration parameter I would prefer leaving it disabled by default and not require it to be able also specified in the config. If someone (really) wants it enabled they can. 
But would like to understand better why making it required may be good idea.

justinstoller: So, IIRC, making it a required key in the two schemas that I mentioned would not make it required in an actual config file. To do that we would need to make it required in the `WebserverRawConfig`.

The two schemas I've mentioned define what's required to our internal handling of the processed config (i.e. they should model what it should look like after your defaults changes have been applied). That means that `ssl-context-factory` could be called without having the `allow-renegotiation` key set. This isn't a huge deal since `nil` is returned by destructuring an unset key in a map [here](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/186/files#diff-1fc7ef9acd5fbc52947b0a4ba167e727R191) and `nil` will count as a `false`-y value when we check [further down](https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/186/files#diff-1fc7ef9acd5fbc52947b0a4ba167e727R216).

Functionally, this creates the same effect, so I'm not going to stress it for now.

PR #187, opened 2017-09-28
(maint) Switch to using `lein` instead of `lein2` in Travis

Magisus:
Travis has stopped including `lein2`, but we don't need to use it
specifically, so this PR updates the Travis script to just use `lein`.

### DIFF
--- a/ext/travisci/test.sh
+++ b/ext/travisci/test.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-lein2 test
+lein test


justinstoller:
For my own edification, this will happen for older projects when Travis enables their trusty environment (now default?) vs their legacy precise environment (lein in [precise](https://docs.travis-ci.com/user/reference/precise/#Leiningen-versions), [trusty](https://docs.travis-ci.com/user/reference/trusty/#Leiningen-version))

PR #189, opened 2018-02-13
(SERVER-2118) Ensure that post request responses can be gzipped

Magisus:
This commit adds POST requests to the list of HTTP methods that can be
gzipped by jetty. By default, only GET requests can request gzipping,
but we fetch catalogs via a post request, and need that response to be
gzipped as well.

### DIFF
--- a/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
+++ b/test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
@@ -50,6 +50,15 @@
         (format "Expected gzipped response, got this response: %s"
                 resp))))
 
+(defn validate-gzip-encoding-when-gzip-requested-on-post-requests
+  [body port]
+  ;; The client/post function asks for compression by default
+  (let [resp (http-sync/post (format "http://localhost:%d/" port))]
+    (is (= (slurp (resp :body)) body))
+    (is (= (get-in resp [:orig-content-encoding]) "gzip")
+        (format "Expected gzipped response, got this response: %s"
+                resp))))
+
 (defn validate-no-gzip-encoding-when-gzip-not-requested
   [body port]
   ;; The client/get function asks for compression by default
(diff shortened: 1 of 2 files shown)


test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
```diff
@@ -50,6 +50,15 @@
         (format "Expected gzipped response, got this response: %s"
                 resp))))
 
+(defn validate-gzip-encoding-when-gzip-requested-on-post-requests
+  [body port]
+  ;; The client/get function asks for compression by default
```
justinstoller: Is this comment still valid?

justinstoller:
The ticket says that there's a change in Jetty that started causing this behavior, I'm curious what was that change?

haus:
@justinstoller I suspect there was no jetty change, but just that we never tested posting catalogs, as most of the other agent/server interaction are over get requests.

justinstoller:
So this isn't so much a bug as a new feature?

haus:
@justinstoller I would call it closing a hole/gap in a previous feature 😄

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_core_test.clj
Magisus: Based on this note in the docs for clj-http-client:
```
:decompress-body: optional; if true, an 'accept-encoding' header with a value of 'gzip, deflate' will be added to the request, and the response will be automatically decompressed if it contains a recognized 'content-encoding' header. Defaults to true.
```
Sounds like it applies to all requests.

haus:
@justinstoller woops, boy was I wrong. I was looking at jetty 9.2.x vs 9.4.x, but we're shipping 9.2.10 in 1.x branch. https://github.com/eclipse/jetty.project/commit/5e3fbbccd0df05049f248e5ac3a9207feaeb3b29 is the change that made it so gzipping was restricted to GET requests by default, and it shipped in 9.2.15, so this was a regression/change in jetty. If we ever move 1.x past 9.2.10 to 9.2.15 or later, we'll need to apply the same fix there.

PR #190, opened 2018-02-14
(maint) Update changelog for 2.1.2

Magisus:


lindboe:
A thing maybe not for this review, but: why was the whitelist necessary in the first place? couldn't this keep coming up?

Magisus:
Apparently defaulting to only gzipping GET requests has a relatively long history in jetty, that predates the creation of the GzipHandler class (its precursor, `GzipFilter`, in Jetty 9.2 doesn't appear to do this, but in earlier versions of 9.4 it did). I'm not sure why, exactly, the commit messages are a little sparse. See @haus 's comment here https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/189#issuecomment-365383656.

PR #198, opened 2019-01-09
(maint) Remove explicit nrepl dependency

robdaemon:
This explicit nrepl dependency is on an older nrepl, and causes conflicts
when users of this library attempt to use the newer nrepl.

Users of this library should specify their own nrepl dependency.

This pairs with puppetlabs/clj-parent#155


robdaemon:
This change is massive. We have tools.nrepl specified all over the place. I'm going to close this for now, and figure out how deep this rabbit hole goes and put a story in JIRA for it.

PR #199, opened 2019-02-21
(PCP-862) Add disconnect function

lucywyman:
This adds a disconnect function to disconnect PCP connections rather than close them, since this will be more performant and close connections more quickly, reducing the number of file descriptors opened at any given time when the connection is flapping.


lucywyman:
It seems the ssl certs may have expired? https://travis-ci.org/puppetlabs/trapperkeeper-webserver-jetty9/builds/496616953#L900

lucywyman:
@MikaelSmith Fixed the ssl certs!

MikaelSmith:
This'll probably need a feature release.

MikaelSmith:
Which means the project.clj needs to be bumped to 2.4.0-SNAPSHOT.

MikaelSmith:
And need CHANGELOG entries (for 2.3.1 and 2.4.0 ideally).

MikaelSmith:
Maybe check with @justinstoller about the release process.

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+## 2.4.0
+
+* [PCP-862](https://tickets.puppetlabs.com/browse/PCP-862) Add disconnect function
+  This adds a disconnect function to disconnect PCP connections rather than
```
MikaelSmith: We probably don't need all this context, but I guess it doesn't hurt.

MikaelSmith: Actually, I don't think we should mention PCP here. I'd just make this a new feature to allow disconnecting websocket connections, it's not specific to PCP.

justinstoller:
Yeah, if you could put in release notes for 2.3.1 that'd be awesome!

Once you're ready to release it should be like any other clj library, I think you just need to run this [jenkins job](https://cinext-jenkinsmaster-enterprise-prod-1.delivery.puppetlabs.net/view/clojure-libraries/view/trapperkeeper-webserver-jetty9/job/enterprise_trapperkeeper-webserver-jetty9_release-clj_master/).

lucywyman:
Ok - I didn't work on 2.3.1 release, so just added the PR commit message to the changelog? It seems like sufficient detail to me :woman_shrugging:

PR #200, opened 2019-02-27
(PCP-862) Only disconnect if the connection is not closed

lucywyman:
When disconnect is called on a connection that is already closed it throws an error. We only want to disconnect if the connection has not already been closed, so this adds an if block to check that `getSession` is not null before disconnecting the session.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,6 +1,6 @@
 (def jetty-version "9.4.11.v20180605")
 
-(defproject puppetlabs/trapperkeeper-webserver-jetty9 "2.5.0-SNAPSHOT"
+(defproject puppetlabs/trapperkeeper-webserver-jetty9 "2.4.1-SNAPSHOT"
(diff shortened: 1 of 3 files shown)


project.clj
```diff
@@ -1,6 +1,6 @@
 (def jetty-version "9.4.11.v20180605")
 
-(defproject puppetlabs/trapperkeeper-webserver-jetty9 "2.5.0-SNAPSHOT"
+(defproject puppetlabs/trapperkeeper-webserver-jetty9 "2.4.0-SNAPSHOT"
```
lucywyman: I'm not sure how this got set too high - I can pull this in to a separate PR, or just leave it if it's not a big deal?

MikaelSmith: Hmm, release job may have been set to bump `:minor` instead of `:patch`. We should change that to 2.4.1-SNAPSHOT, which comes before 2.4.1.

lucywyman:
pcp-tests
This did also make all the pcp-broker tests pass!

PR #201, opened 2019-05-24
(maint) new jetty

justinstoller:
This resolves issues when upgrading from jetty 9.4.11 to 9.4.18, however it involves disabling some previously used "weak" ciphers and removing an option that was abruptly ignored by Jetty. An important change is the setting of endpointIdentification, which we do instead of using `SslContextFactory.Server`.


src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
```diff
@@ -197,6 +197,7 @@
                   (.setKeyStore (:keystore keystore-config))
                   (.setKeyStorePassword (:key-password keystore-config))
                   (.setTrustStore (:truststore keystore-config))
+                  (.setEndpointIdentificationAlgorithm nil)
```
justinstoller: So, Jetty's preferred way to handle this issue is actually to create an instance of `SslContextFactory.Server` (which is a static inner class that will have the correct configuration).

However we are currently using `InternalSslContextFactory` which is our _own_ subclass, that handles how we reload the CRL.

In Jetty 10, SslContextFactory is becoming an abstract class and its methods are currently deprecated. I'm unsure if now is the correct time to rework that subclassing, and if so, the best way to go forward.

justinstoller:
Looks like I need to update the travis config in addition to re-issuing the artifacts used in the crl test.

justinstoller:
@jonathannewman @rbrw , this might interest you.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -58,19 +58,9 @@
...
+  ["TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256"
    "TLS_DHE_RSA_WITH_AES_256_CBC_SHA256"
-   "TLS_DHE_RSA_WITH_AES_256_CBC_SHA"
-   "TLS_DHE_RSA_WITH_AES_128_CBC_SHA256"
-   "TLS_DHE_RSA_WITH_AES_128_CBC_SHA"
-
-   "TLS_RSA_WITH_AES_256_CBC_SHA256"
-   "TLS_RSA_WITH_AES_256_CBC_SHA"
-   "TLS_RSA_WITH_AES_128_CBC_SHA256"
-   "TLS_RSA_WITH_AES_128_CBC_SHA"])
```
justinstoller: This should probably add a bunch more suites

jonathannewman:
Thanks for doing this.  It looks like Travis is complaining that jdk8 is too old (if I'm interpreting the error correctly).  Maybe the travis config needs to be updated too?

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
justinstoller: I added all of the cipher suites enabled by default on my local box using Java 11 that Jetty didn't warn about in a tongue-in-cheek commit at 3074952. I figured that at least might start a conversation about how to figure out what cipher suites we should be using.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
justinstoller: I looked into this a little bit more and didn't see an obvious way forward besides what I'm doing here (but I'm not a Java expert and would love someone to double check me on this).

The concern about changes to Jetty 10 are probably moot considering this is "trapperkeeper-webserver-jetty _9_ ".

justinstoller:
I'll put "(maint) Update travis to test openjdk vs oraclejdk" (0bd321d) through "(maint) Use a more specific List to avoid unchecked warnings" (24872af) into a separate PR.

I think there's still some discussion to have here wrt which cipher suites should be available.

I'm also a little stumped on the remaining test failure. I thought it was an expired crl but it's actually the crl that Lucy updated in February (worked then and it and the associated ca is valid for another decade). As I mentioned earlier, it also fails for me on master.

justinstoller:
I've opened https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/202 with the fixes for master. Once that's in I'll open another based on it with the Java 11 updates I accidentally did. And rebase those changes out of this PR.

justinstoller:
Java 11 support is at https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/203 so it can be considered separately. I think is read for meaningful review.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -576,7 +576,7 @@
             (is (nil? second-jetty-server)
                 "Jetty server was unexpectedly attached to the service context")
             (is (thrown?
-                  BindException
+                  IOException
```
rbrw: Suggest adding a bit to the commit message explaining the motivation for the change.  And not sure yet, but this might affect pdb (though possibly just our tests).

rbrw: (For pdb, I suspect it'll be a trivial adjustment, if we need one, but wondered if it might affect any other things.)

doc/jetty-config.md
```diff
@@ -94,13 +94,16 @@ header is being sent in an HTTP Request.
...
+This setting has been removed. The option used in Jetty no longer attempts to
+set the SO_LINGER for the socket as of v9.4.12.
+
+The Jetty maintainers discovered this option has undefined behavior on
+non-blocking connections (and all underlying connections became non-blocking
+in Jetty 9.0).
+
+The actual behavior, though undefined by the Java spec, also changes between
+Java versions (so a user on Java 8 will see different behavior on Java 11) and
+has been reported to the Jetty community as a source of bugs on some platforms.
```
rbrw: Wondered if this might suggest an X release, i.e. 10 rather than 9, assuming we consider it a backward incompatibility.

rbrw: Oh wait, I guess that only makes sense if the tk-jetty-9 versions are unrelated to the underlying jetty version, and I don't know what was originally intended either way on that front.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
rbrw: Suggest maybe adding a bit of explanation about the rationale/effect of this change to the commit message.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
robdaemon: We should compare this to a valid cipher suite list for FIPS compliance

src/puppetlabs/trapperkeeper/services/webserver/jetty9_core.clj
robdaemon: `setEndpointIdentificationAlgorithm` is a client-side thing to validate that the certificate matches the host that sent it (X509v3 extensions, etc) - it should be safe to make this nil in a server.

As far as reloading the CRL, I guess we get to rework our class when Jetty 10 happens.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
```diff
@@ -942,8 +942,10 @@
       app
       [jetty9-service
        hello-webservice]
-      (assoc-in jetty-ssl-pem-config [:webserver :ssl-protocols] ["SSLv3"])
-       (let [response (http-get "https://localhost:8081/hi_world" (merge default-options-for-https-client
-                                                                         {:ssl-protocols ["SSLv3"]}))]
-         (is (= (:status response) 200))
-         (is (= (:body response) "Hi World")))))))
+      (-> jetty-ssl-pem-config
```
robdaemon: Shouldn't this test only verify that SSLv3 is permanently disabled? We should not allow SSLv3 to be used ever.

doc/jetty-config.md
justinstoller: I think since this might break consumers of tk-ws-jetty9 it warrants an X bump to 3.0 for the library.

test/clj/puppetlabs/trapperkeeper/services/webserver/jetty9_service_test.clj
justinstoller: I think this is just testing the configurability of the setting in general. It could also be tested by using the setting to disable TLSv1.1 as well. (or at least that's what I got from the PR/ticket: https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/92/files)

robdaemon: Ahhh okay, sorry! Still getting the lay of the land with the other parts of the stack.

justinstoller:
fwiw, I just ran the Puppet Server tests with this merged into the Java 11 PR on Java 11, and saw two errors and some additional output in a non-failing test to the logs. Both errors came from the restriction of default ciphers, I'm unclear on the source of the additional output to the logs.

Server uses this library extensively in its tests so I consider those minimal issues a positive signal.

justinstoller:
I will update the commit messages where needed and look into using the ciphers supported by fips level 2 over the next couple of days.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
justinstoller: I've pushed up the list of enabled cipher suites on a RHEL7 fips enabled VM you got @robdaemon (7027e2c). I then removed any that Jetty also warned were weak (4caf517). I also diffed those against the cipher suites in the PR previously that I had gotten from Java in OSX Mojave. Here is that diff:
```
tk-jetty9 justin (newjetty %) ‖ wc -l osx
      25 osx

tk-jetty9 justin (newjetty %) ‖ wc -l fips
      33 fips

tk-jetty9 justin (newjetty %) ‖ diff osx fips
1,2c1
< TLS_AES_128_GCM_SHA256
< TLS_AES_256_GCM_SHA384
---
> SSL_CK_DES_192_EDE3_CBC_WITH_SHA
11c10,17
< TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256
---
> TLS_DH_DSS_WITH_AES_128_CBC_SHA256
> TLS_DH_DSS_WITH_AES_128_GCM_SHA256
> TLS_DH_DSS_WITH_AES_256_CBC_SHA256
> TLS_DH_DSS_WITH_AES_256_GCM_SHA384
> TLS_DH_RSA_WITH_AES_128_CBC_SHA256
> TLS_DH_RSA_WITH_AES_128_GCM_SHA256
> TLS_DH_RSA_WITH_AES_256_CBC_SHA256
> TLS_DH_RSA_WITH_AES_256_GCM_SHA384
21a28
> TLS_ECDH_ECDSA_WITH_AES_256_GCM_SHA384
25a33
> TLS_SRP_SHA_DSS_WITH_AES_256_CBC_SHA
```

In other words osx is a subset of the fips suites save for three that RHEL7 didn't have: `TLS_AES_128_GCM_SHA256`, `TLS_AES_256_GCM_SHA384`, and `TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256`

The first two look to be TLSv1.3. (see https://crypto.stackexchange.com/questions/62917/what-is-different-below-two-ciphersuites). The third looks older though potentially still supported? I assume we can leave off all three for the time being?

justinstoller: If the current list of cipher suites looks sane, I'll go ahead and rebase those commits away and update the other commit messages and I think we'll be good to go.

robdaemon: Looks sane to me :+1:

justinstoller:
I alphabetized the list of cipher suites, rebased, and improved commit messages. Locally, the "refresh cipher suites" commit follows directly after the "remove weak ciphers" commit as it logically should. I'm not sure what's happening there. Regardless, I think this is read to go.

PR #202, opened 2019-05-28
(maint) Fix master

justinstoller:
When doing other work I've found that master is broken because of a couple of expired CRLs (note some other similar artifacts were already regenerated here: https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/199).

This updates us to use OpenJDK8 instead of OracleJDK8 because it seems that OracleJDK8 is no longer a valid version in travis.


jonathannewman:
LGTM! Expired?

jonathannewman:
merge party

PR #203, opened 2019-05-28
(maint) Support Java 11 & update dependencies

justinstoller:


justinstoller:
fwiw, I've also tested these changes locally in a snapshot build with Puppet Server master (vs Java8 compiled and executed, Java8 compiled and executed on Java11, and Java11 compiled and executed).

PR #207, opened 2019-06-26
(maint) Suppress warning about empty contextPath

pcarlisle:
Jetty warns if a contextPath is an empty string. Since that is normal in
clients of this library, substitute "/" in this case, which is what
jetty ends up using anyway.


pcarlisle:
Maybe we need a 2.x branch

PR #210, opened 2019-10-01
(PE-26658) apply JVM security policy to FIPS

robdaemon:
Add a FIPS profile, with a JVM security profile to limit the crypto
providers in an FIPS environment. Fixes several tests (and core
functions) to work in FIPS.


robdaemon:
Tests in FIPS mode do not work in jdk11. I'm going to take the next 24 hours to resolve that, just to not need to take this up again. Please don't merge this yet.

project.clj
```diff
@@ -37,17 +37,19 @@
                  [ring/ring-servlet]
                  [ring/ring-codec]
 
-                 [puppetlabs/ssl-utils]
+                 [puppetlabs/ssl-utils "3.0.4"]
```
robdaemon: this is temporary, we need a new clj-parent release

project.clj
```diff
@@ -37,17 +37,19 @@
...
                  [puppetlabs/kitchensink]
                  [puppetlabs/trapperkeeper]
                  [puppetlabs/i18n]
                  [puppetlabs/trapperkeeper-filesystem-watcher]
+
+                 [org.slf4j/jul-to-slf4j "1.7.20"]
```
robdaemon: need to add this into clj-parent

PR #212, opened 2019-10-16
(maint) fixing misconfiguration causing noise

robdaemon:
Fixing some wiring to the SslContextFactory in FIPS mode that was
causing jetty to be extra noisy. Specifically, this will stop these log
messages from occurring:

    2019-10-16 12:04:50,761 INFO  [o.e.j.u.s.SslContextFactory] No Cipher matching 'TLS_DHE_RSA_WITH_AES_128_GCM_SHA256' is supported
    2019-10-16 12:04:50,761 INFO  [o.e.j.u.s.SslContextFactory] No Cipher matching 'TLS_DHE_RSA_WITH_AES_256_GCM_SHA384' is supported
    2019-10-16 12:04:50,761 INFO  [o.e.j.u.s.SslContextFactory] No Cipher matching 'TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384' is supported
    2019-10-16 12:04:52,769 INFO  [o.e.j.u.s.SslContextFactory] Unable to get CertificateFactory instance for type [X.509] on provider [BCJSSE], using default

Note: this involved removing three supported ciphers from the default
list, so that will require additional testing. Downstream consumers of
tk-jetty9 will also want to update their configured cipher lists in FIPS
mode.

### DIFF
--- a/dev-resources/logback.xml
+++ b/dev-resources/logback.xml
@@ -6,9 +6,9 @@
     </appender>
 
     <logger name="org.apache.http" level="warn"/>
-    <logger name="org.eclipse.jetty" level="warn"/>
+    <logger name="org.eclipse.jetty" level="info"/>
 
-    <root level="warn">
+    <root level="info">
         <appender-ref ref="STDOUT" />
     </root>
 </configuration>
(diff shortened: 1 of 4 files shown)


justinstoller:
I don't have any thoughts on the cipher suite changes, though someone from the installer might.

Excited about the logging fix. This will need releasing in clj-parent and all the PE components will need to take up that version of clj-parent.  /cc @puppetlabs/puppetserver-maintainers @jonathannewman

robdaemon:
Yeah, on the cipher suites change, I removed them because the logging was saying `No Cipher matching '...' is supported`

Knowing that we override the cipher lists in puppetserver and pdb (or is that only for FOSS?) we might see more messages. Another thing to look at the logs of an installed FIPS environment for.

PR #213, opened 2020-02-13
(SERVER-2734) Disambiguate overloaded method call

Magisus:
This commit adds a type annotation to indicate which version of Jetty's
`newResource` function is desired.


Magisus:
The bug related to this is showing up in agent PEZ now too. We'll need to get this into clj-parent and get it into all three server projects.

PR #216, opened 2020-04-02
Upgrade to latest jetty

svartulfr:
Uses latest (9.4.27.v20200227) version of Jetty.  This version,
however, breaks tests because it incorrectly decodes overlong UTF-8
text (specifically normalize-uri-with-overlong-utf8-chars-tests).  To
avoid potential security issues the 9.4.18 version of
URIUtil.decodePath has been copy/pasted.


svartulfr:
Build failed retrieving leiningen -- I'd like to restart, but don't have ability to do that in Travis.

Magisus:
Looks like we've updated to a newer version of jetty since this was opened. Thanks for putting this up, but since it's been taken care of, going to close this.

PR #218, opened 2020-05-18
(maint) Update changelog for 4.1.0

Magisus:


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 4.1.0
```
mwaggett, line 1: I assume we're calling this a Y because Jetty is the main dependency in this library?

Magisus: Yeah I think so.

justinstoller: Yeah, and Jetty doesn't really follow semver. I think this accurately reflects the risk involved with this update.

PR #219, opened 2020-09-23
Make the WebserverService protocol and service be implemented via metadata-based protocol extension

vemv:


vemv:
Sorry, this PR wasn't meant to target this org.

PR #222, opened 2021-01-27
(maint) Update flatland to 1.5.9

Magisus:
This commit updates flatland to latest, to match what's pulled in by
other libraries, notably clj-yaml.


Magisus:
This dep is not in clj-parent, so we have to specify the version explicitly.

steveax (approved):
👍

austb (approved):

Magisus:
I should get the changelog update in here.

PR #225, opened 2021-04-14
(SEC-431) update jetty to 9.4.39.v20210325 to resolve CVEs

steveax:
- CVE-2021-28165 - #6072 - jetty server high CPU when client send data length > 17408
- CVE-2021-28164 - #6101 - Normalize ambiguous URIs
- CVE-2021-28163 - #6102 - Exclude webapps directory from deployment scan

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -1,4 +1,4 @@
-(def jetty-version "9.4.36.v20210114")
+(def jetty-version "9.4.39.v20210325")
 
 (defproject puppetlabs/trapperkeeper-webserver-jetty9 "4.1.5-SNAPSHOT"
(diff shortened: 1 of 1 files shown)


mwaggett:
Does this need to land in the imminent PE releases? Is it okay that this is gonna miss the Platform releases?

steveax:
@mwaggett, not sure, I just noticed these when I ran lein-nvd. Probably should check with @pup-shellee and some other folks but I thought I'd get the ball rolling in any event.

mwaggett (approved):

austb (approved):

mwaggett:
folks seemed to agree that we should just get this in and let jenkins do the testing, so merging

PR #231, opened 2021-08-31
(TK-494) Enable TLSv1.3 by default

mwaggett:
see individual commit messages for details


mwaggett:
Running puppetserver#6.x tests with this right now.

mwaggett:
oh good, it doesn't like dots in the function names

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -58,20 +58,28 @@
...
 ;;; keep these overrides.
-(def acceptable-ciphers
-  ["TLS_DHE_RSA_WITH_AES_128_GCM_SHA256"
-   "TLS_DHE_RSA_WITH_AES_256_GCM_SHA384"
-   "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256"
-   "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384"
+(def tls-1-dot-3-ciphers
+  ["TLS_AES_128_GCM_SHA256"
+   "TLS_AES_256_GCM_SHA384"
+   "TLS_CHACHA20_POLY1305_SHA256"])
```
justinstoller, line 64: what'd Shellee say about these three vs the five listed in the ticket?

mwaggett: she hasn't answered yet

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -58,20 +58,28 @@
...
+
+(def tls-1-dot-2-ciphers
+  ["TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256"
```
justinstoller, line 67: I think these are in the wrong order?

mwaggett: oh, I was looking at Cipher Suite Preferences, point #2 says "The order of cipher suites to be used by preference is:", so I followed that. But I see that that doesn't exactly match the Preferred Cipher Suite Order table.. I'll ask Shellee about it.

justinstoller: I think that's saying we should prefer all the 1.3 ciphers before the 1.2 ciphers, then below she lists the order that 1.2 ciphers should be in.

mwaggett: yeah, I think you're right, but I'm gonna confirm with Shellee (who is apparently on PTO this week), since it's confusing.

mwaggett: okay, looks like what I have is right. According to Diego, we're following Mozilla's recommendations: https://wiki.mozilla.org/Security/Server_Side_TLS

mwaggett:
puppetserver#6.x tests passed with this.

@justinstoller How important do you think it is to investigate adding the `CHACHA20` ciphers for TLS 1.2? If it is important, what do you think of us doing that as a separate PR (and/or ticket)?

justinstoller:
I think it's fine to do it as a separate ticket, prioritized independently

mwaggett:
cool, I'll open that ticket.

This PR should be good to go then, unless I'm missing something.

PR #232, opened 2021-09-16
(maint) Add TLSv1.3 ciphers to FIPS cipher list

justinstoller:


src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -63,21 +63,21 @@
...
-   "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384"
+   "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384"])
+
+(def acceptable-ciphers-fips
+  (into [] (concat tls-1-dot-3-ciphers fips-tls-1-dot-2-ciphers)))
+
+(def non-fips-tls-1-dot-2-ciphers
+  ["TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384"
    "TLS_DHE_RSA_WITH_AES_128_GCM_SHA256"
    "TLS_DHE_RSA_WITH_AES_256_GCM_SHA384"])
```
mwaggett, line 77: sooo I'm kinda thinking we should keep the `tls-1-dot-2-ciphers` list in addition to the `fips-tls-1-dot-2-ciphers` list you just added, even though it means listing some ciphers in both lists. This is one of the things I was thinking about when originally doing this work and went back and forth about, but I think it's ultimately clearer to have two separate lists. By having a `non-fips` list and then `concat`ing it with the `fips` list, we don't have the freedom to arrange the cipher lists in our preferred order - I don't know if we can always guarantee that the FIPS ciphers will always be preferred over the others in non-FIPS environments, even though that happens to be the case right now. It also looked wrong at first glance to see `acceptable-ciphers-fips` in the non-FIPS `acceptable-ciphers`, even though I can see now why that works.

justinstoller: updated

mwaggett (approved):

mwaggett:
ahh @justinstoller this should probably get a CHANGELOG entry

mwaggett (approved):

PR #233, opened 2021-10-04
(PE-32764) Update default ciphers

nmburgan:
This makes several changes to the default cipher lists for Jetty.
- Removes the TLS_CHACHA20_POLY1305_SHA256 cipher, as it is currently unsupported in Java 11.
- Adds TLS_DHE_RSA ciphers as recommended by the Mozilla Intermediate list.
- Adds the RSA version of the ECDHE AES 256 cipher to the FIPS list. Since our CA issues RSA-type certs/keys, we effectively only had one useable cipher.
- Rearranges the cipher list to be in the preferred order.


src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -60,24 +60,24 @@
...
+   "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"
+   "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256"
+   "TLS_DHE_RSA_WITH_AES_256_GCM_SHA384"
+   "TLS_DHE_RSA_WITH_AES_128_GCM_SHA256"])
 
 (def acceptable-ciphers
   (into [] (concat tls-1-dot-3-ciphers tls-1-dot-2-ciphers)))
 
 (def fips-tls-1-dot-2-ciphers
-  ["TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256"
```
justinstoller, line 78: I thought this work was to remove ChaCha20, what's up with the rest of the changes? (should also be in the commit message)

nmburgan: Sorry, you're right, I rushed through this one.  Let me add some notes.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
```diff
@@ -60,24 +60,24 @@
...
 (def tls-1-dot-2-ciphers
-  ["TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256"
-   "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"
+  ["TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384"
    "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384"
-   "TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384"
-   "TLS_DHE_RSA_WITH_AES_128_GCM_SHA256"
-   "TLS_DHE_RSA_WITH_AES_256_GCM_SHA384"])
+   "TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"
+   "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256"
```
justinstoller, line 69: So I think this is a better ordering of the suites, but I'm going to pull @mwaggett in on this because she did the leg work on why they are in the order they are now (see https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/231)

justinstoller:
wrt the addition of some ciphers, I thought we were only using those allowed by an internal cipher list policy (one which may be stricter than the mozilla intermediate list, since we're often slower than they are to update).

nmburgan:
The ciphers were actually pulled from https://confluence.puppetlabs.com/display/SRE/TLS+Usage+Standard but I think that was generated off the intermediate list.

mwaggett:
For the future, it would be great if changes like this could be broken up into smaller commits, as small as the changes are. That makes it easier to look at each of the individual changes on their own and makes it easier to revert just one piece if we need to.

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
nmburgan: The site doesn't really have the correct ordering.  I double checked with Jeremy on what the ordering should be and https://github.com/puppetlabs/puppet-enterprise-modules/blob/main/modules/puppet_enterprise/functions/ciphers.pp is the right order.  Generally, we want 256 before 128, and certain new ciphers before other older ones.

mwaggett: ahkay, well in that case, could you or @LivingInSyn update https://confluence.puppetlabs.com/display/SRE/TLS+Usage+Standard to match (and remove references to Mozilla's ordering if that's not accurate/what we're using)?

mwaggett (approved):

src/puppetlabs/trapperkeeper/services/webserver/jetty9_config.clj
nmburgan: Just updated, waiting on final check by @LivingInSyn to make sure I didn't say anything incorrect.

mwaggett: Looks like the Mozilla references are still in there. I can remove them and @LivingInSyn can check that too. 😁

nmburgan: It's still based on the Mozilla Intermediate list.  Just, for some reason, the Mozilla wiki lists them in a weird order with 128s before 256s.

nmburgan: I'm also not 100% sure if order actually matters or is followed by the various services, and I haven't looked into a way to test that.

mwaggett: Jake Peralta from Brooklyn Nine-Nine saying 'cool' a lot

PR #241, opened 2022-09-30
(maint) update clj-parent to 5.2.6

jonathannewman:
This updates clj-parent to 5.2.6 to include the new version of bouncy-castle which has been renamed from `15on` to `18on` to indicate that it is now only jdk8 and beyond compatible.

Additionally the changelog was updated to prepare for a new 4.4.0 release.


steveax (approved):
👍

PR #242, opened 2022-10-02
(PE-34658) Bump clj-parent to remove stylefruits/gniazdo version

mwaggett:
This updates the clj-parent version to 5.2.9, which contains a version declaration for stylefruits/gniazdo, allowing the version in this repository to not pin a specific version.


jonathannewman:
@mwaggett  good catch!

jonathannewman (approved):

jonathannewman:
I think this means yet another clj-parent bump too

mwaggett:
yeah, probably should

PR #243, opened 2022-10-02
(maint) remove dependency on stylefruits/gniazdo

jonathannewman:
Based on a suggestion in the gniazdo project https://github.com/stalefruits/gniazdo#usage this was updated to use https://github.com/gnarroway/hato instead of stylefruits, which requires it to drop support of jdk8 for testing.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -76,15 +76,13 @@
                                        [puppetlabs/trapperkeeper nil :classifier "test"]
                                        [org.clojure/tools.namespace]
                                        [compojure]
-                                       [stylefruits/gniazdo nil :exclusions [org.eclipse.jetty.websocket/websocket-api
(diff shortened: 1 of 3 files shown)


steveax:
I'm confused about how this and #244 play together.

jonathannewman:
It is an alternative, and eliminates jdk8 testing.  Seemed like #244 would be safer to introduce earlier.

mwaggett:
We can't drop support for jdk8 in Puppet 7, so idk if we want this right now. Maybe something to ticket and come back to for Puppet 8?

jonathannewman:
It technically doesn't drop support for jdk8, just drops testing of this project in jdk8.

jonathannewman:
And yes, it is something we should revisit.

PR #248, opened 2023-04-20
(maint) restore jdk 8 compatiblility

jonathannewman:
This library must be built against jdk8 for FOSS compatiblity


steveax (approved):
👍

PR #250, opened 2023-05-17
(PE-35600) Provide response object to handler when :include-response?

rbrw:
This is just one way we might solve the problem, and the *pdb-response* binding is just there to support the original investigation, without having to rework pdb's query handlers to accept the additional argument.  Presumably we'll remove that if we decide to pursue these changes.


rbrw:
Continued as https://github.com/puppetlabs/trapperkeeper-webserver-jetty9/pull/251

PR #251, opened 2023-05-31
(PDB-5645) Call handler with response when :include-response set

rbrw:
If :include-response is true in an add-ring-handler call, then call the handler with the pending request Response as a second argument so that the handler can interrogate/manipulate the response directly if desired.  PuppetDB intends to use this to detect client disconnects via the SocketChannel (via the Response getHttpChannel) in order to halt a potentially expensive query sooner.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,11 @@
+## 4.5.0
+
+* Add `:include-response` option to request that the pending response
+  instance be provided to
+  [`add-ring-handler`](./README.md#add-ring-handler) as a second
+  argument.
+  [(PDB-5645)](https://tickets.puppetlabs.com/browse/PDB-5645)
+
 ## 4.4.3
 * restore jdk 8 compatibliity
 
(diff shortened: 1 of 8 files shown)


rbrw:
I added the "don't merge" since this will support a forthcoming PDB pr, and thought I'd like to get that one settled too before we proceed, though I don't currently anticipate any further changes to this side.

jonathannewman:
@rbrw Note that we are working on the jetty 10 version of this currently (we have everything but the web socket implementation working)  with the intent of putting it into all the main branches soon.

jonathannewman:
See https://github.com/puppetlabs/trapperkeeper-webserver-jetty10 for reference.

rbrw:
OK, thanks.  Appreciate the notice.  I think we'd like to merge this if we can sometime in the next two weeks (though after the releases), and yep, it's currently only intended for pdb 8.

Also had to rework the approach a bit (now done).

PR #253, opened 2023-10-10
(PE-36979) update to jetty 9.4.53.v20231009 to address two CVEs / prepare for release

jonathannewman:
See individual commit comments.


steveax (approved):
👍

steveax:
Not sure what's up with the PR tests, but ran them locally 👍

PR #254, opened 2023-10-10
(maint) make service test less sensitive

jonathannewman:
The service test contained a regular expression that matched on the size of the content. The headers that are included in the response depend on the platform and java version that are used to run the tests, so this removes the size part of the regular expression to make the tests less sensitive to environment. The underlying behavior in the test isn't impacted; it still validates that the logging occurred.


steveax (approved):
👍

